PR #7, opened 2016-07-22
Add Docker configuration so the service can be run in a Docker container

JoeChapman:
- add Dockerfile
- add docker-compose.yml
- add run.sh

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,32 @@
+version: "2"
+services:
+  app:
+    build: .
+    environment:
+      - NODE_ENV=docker-compose
+      - REDIS_HOST=redis
+      - REDIS_PORT=6379
+    links:
+      - redis
+    ports:
+      - "8080:8080"
+    volumes:
+      - "/public"
+  redis:
+    image: redis
+
+  nginx-proxy:
+    image: quay.io/ukhomeofficedigital/nginx-proxy:v1.5.1
+    environment:
+      - PROXY_SERVICE_HOST=app
+      - PROXY_SERVICE_PORT=8080
+      - NAXSI_USE_DEFAULT_RULES=FALSE
+      - ADD_NGINX_SERVER_CFG=add_header Cache-Control private;add_header X-Frame-Options "SAMEORIGIN" always;add_header X-Content-Type-Options "nosniff" always;add_header X-XSS-Protection "1; mode=block" always;location /public {add_header Cache-Control max-age=86400;add_header X-Frame-Options "SAMEORIGIN" always;add_header X-Content-Type-Options "nosniff" always;add_header X-XSS-Protection "1; mode=block" always;alias /public;}
+      - ERROR_REDIRECT_CODES=599
+    ports:
+      - "443:443"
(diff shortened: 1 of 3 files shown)


run.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+if [ "$NODE_ENV" = "ci" ]
+then echo "starting service"
+  SITEROOT=/firearms GA_TAG_ID=UA-72527484-1
```
easternbloc: is that the correct GA TAG?

docker-compose.yml
```diff
@@ -0,0 +1,32 @@
+version: "2"
```
bendiggle: Should it be v2?

docker-compose.yml
```diff
@@ -0,0 +1,32 @@
...
+services:
```
bendiggle: Also, is firearms not going to need maildev?

easternbloc: Probably not. At least not right now.

docker-compose.yml
easternbloc: Yes, We've migrated to compose v2

run.sh
JoeChapman: TBH, I am not sure what SITEROOT is for. Does anyone lnow if it's still required? @easternbloc @gxxm @joefitter @bendiggle ???

easternbloc: it's used to make assetPath which is then uses across our templates.

easternbloc: Not sure why it's prefixed with the app name though. BRP doesn't do this...

easternbloc: Sadly bad commit messages do not explain it: https://github.com/UKHomeOffice/gro/commit/069eca1f

JoeChapman: Looks like BS to me. Why would baseUrl ever need to be `firearms/new-dealer` or `brp/not-arrived` or such like? It's as if the app is serving from a level above root, it's wrong.

JoeChapman:
@easternbloc @bendiggle?

run.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+if [ "$NODE_ENV" = "ci" ]
+then echo "starting service"
+  GA_TAG_ID=
```
easternbloc: have you created a ticket for analytics so this doesn't get missed...

docker-compose.yml
```diff
@@ -0,0 +1,32 @@
...
+      - redis
+    ports:
+      - "8080:8080"
+    volumes:
+      - "/public"
+  redis:
+    image: redis
+
+  nginx-proxy:
+    image: quay.io/ukhomeofficedigital/nginx-proxy:v1.2.3
```
easternbloc: the latest nginx-proxy is 1.5.1 I'd suggest updating to that as it's a fresh project

run.sh
JoeChapman: Yes: https://jira.digital.homeoffice.gov.uk/browse/FLHO-93

bendiggle:
LGTM

PR #8, opened 2016-08-15
Set up drone CI config

JoeChapman:
- remove travis config
- update dotfiles
- add Dockerfile for acceptance tests
- add test shell script and make exectuable


drone.yml
```diff
@@ -0,0 +1,46 @@
+pipeline:
```
easternbloc: needs more more constrains for when to build the pipelines. SEE latest version of /brp_app

PR #9, opened 2016-09-09
Feature/handle

easternbloc:
Added the handle step to the firearms form
- Added the drone file and sig file,
- Updated npm to 3.10.8 to Dockerfile to prevent memory leak
- Updated Dockerfile-acceptance with codecept config and apps directory,
- Added feature acceptance file for handle step,
- Added handle page helper,
- Added obtain page helper,
- Updated fields config with 2 checkboxes which are dependant on each other,
- Added the handle step to the routes config,
- Added the handle step field translations,
- Added the title translations file, 
- Added the pages translation file with headers in them,
- Added validation translation file for custom validation messages,
- Added codecept config file,
- Update package.json with bootstrap development branch, so-acceptance and a few changes to the npm run scripts


.eslintignore
```diff
@@ -0,0 +1,9 @@
+deploy
```
JoeChapman: do you need this?

.eslintignore
```diff
@@ -0,0 +1,9 @@
...
+public
+secrets.env
```
JoeChapman: do you need this?

.eslintrc
```diff
@@ -0,0 +1,5 @@
+extends:
+- "homeoffice/config/default"
+
+env:
+  es6: true
```
JoeChapman: I think this is unnecessary for the version of eslint we use

bendiggle: nope this is still needed

PR #15, opened 2016-09-22
Feature/flho146 obtain Step

bendiggle:
* Added the step to the routes config,
* Added the translations to the fields pages and validation files. This includes deepTranslations,
* Added a template partial for the buy checkbox toggle,
* Added a line of sass for the partial removing a bottom margin,
* Added the acceptance tests for the obtain step,
* Added the page helper for obtain and storage,
* Added the helpers to the codecept config file.

### DIFF
--- a/apps/new-dealer/acceptance/pages/obtain.js
+++ b/apps/new-dealer/acceptance/pages/obtain.js
@@ -1,5 +1,100 @@
 'use strict';
 
+const TRANSLATIONS = require('../../translations/en/default');
+const FIELDS = TRANSLATIONS.fields;
+const PAGES = TRANSLATIONS.pages;
+
+let I;
+
 module.exports = {
-  url: 'obtain'
+
+  _init() {
+    I = require('so-acceptance/steps.js')();
+  },
+
+  url: 'obtain',
+  'obtain-group': '#obtain-group',
+  buy: '#obtain-buy',
+  'temporary-possession': '#obtain-temporary-possession',
+  manufacture: '#obtain-manufacture',
+  'other-means': '#obtain-other-means',
+  'wont-take-possession': '#obtain-wont-take-possession',
+
+  'further-details': {
+    buy: '#buy-details',
+    'temporary-possession': '#temporary-details',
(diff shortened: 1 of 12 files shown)


joefitter:
💩

bendiggle:
@joefitter I'm going to take that as a LGTM!

joefitter:
I want to make some changes to the deepTranslate which will affect this step, but, let's merge, bcoz agile

joefitter:
@JoeChapman @gxxm

bendiggle:
We're so agile!

apps/new-dealer/acceptance/pages/obtain.js
```diff
@@ -1,5 +1,86 @@
 'use strict';
 
+const TRANSLATIONS = require('../../translations/en/default');
+const FIELDS = TRANSLATIONS.fields;
+const PAGES = TRANSLATIONS.pages;
```
JoeChapman: Is this how we do? capitalise constants?

apps/new-dealer/fields/index.js
```diff
@@ -3,7 +3,7 @@
 module.exports = {
   company: {
     mixin: 'radio-group',
-    validate: ['required'],
+    validate: 'required',
     className: ['form-group'],
```
JoeChapman: DO single className's need to be in array notation?

joefitter: either or

joefitter: string is nicer if only one value - we save 2 bytes ;)

bendiggle: Yep should be outside of the array. But doesn't link to this PR. Next one!

apps/new-dealer/acceptance/pages/obtain.js
joefitter: I like this pattern

bendiggle: I like this pattern too. We've been doing it on the other forms so just been consistent here

PR #25, opened 2016-10-10
flho-165 Authority-holders-address

bendiggle:
Authority holders address. This has the address steps for the first and second authority-holders and consists of postcode step, manual-address step and address-lookup step.

	* Added the steps to the route config,
	* Added the controllers for the postcode, address and address lookup,
	* Added the custom html for postcode, address and address lookup,
	* Added the validation translations for postcode, address and address lookup,
	* Added the page translations for postcode, address and address lookup,
	* Added the field translations for postcode, address and address lookup,
	* Edited the config to use the postcode lookup api,
	* Added the mock-postcode for the acceptance tests,
	* Edited the app.js to use the mock-postcode when env var is ci,
	* Added the acceptance tests for both the first and second authority holders,
	* Added the step helped for the acceptance tests.


apps/new-dealer/controllers/address-lookup.js
```diff
@@ -0,0 +1,49 @@
...
+    if (req.form.values[key] === this.options.fields[`${field}-address-lookup`].options[0].value) {
+      return new ErrorController(`${field}-address-lookup`, {
+        key: `${field}-address-lookup`,
+        type: 'required',
+        redirect: undefined
+      });
+    }
+    return;
+  }
+};
```
JoeChapman: I'm feeling an AddressLookup Controller being added to HOF controllers. But maybe not yet

apps/new-dealer/controllers/address-lookup.js
```diff
@@ -0,0 +1,49 @@
...
+  validateField(key, req) {
+    const field = this.options.locals.field;
+    if (req.form.values[key] === this.options.fields[`${field}-address-lookup`].options[0].value) {
+      return new ErrorController(`${field}-address-lookup`, {
+        key: `${field}-address-lookup`,
+        type: 'required',
+        redirect: undefined
+      });
+    }
+    return;
```
JoeChapman: This is unnecessary, a function without a return statement will implicitly return undefined

apps/new-dealer/controllers/address-lookup.js
bendiggle: Yeah, I think so. Sooner rather than later.

apps/new-dealer/controllers/address-lookup.js
bendiggle: It has been removed

PR #30, opened 2016-10-18
flho-163 Storage-address section

bendiggle:
storage-address section which includes postcode, address lookup and manual address steps.

	* Added the steps to the route config,
	* Added the translations for fields, pages and validation
	* Added the acceptance tests for the steps

### DIFF
--- a/apps/new-dealer/index.js
+++ b/apps/new-dealer/index.js
@@ -85,7 +85,7 @@ module.exports = {
           return stored && weaponsAndAmmo;
         }
       }, {
-        target: '/storage-address',
+        target: '/storage-postcode',
         condition(req) {
           const stored = storedOnPremises(req);
           const weaponsAndAmmo = weapons(req) && ammunition(req);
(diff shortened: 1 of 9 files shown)


apps/new-dealer/index.js
```diff
@@ -103,12 +103,59 @@ module.exports = {
...
+        field: 'storage'
+      }
+    },
     '/storage-address': {
+      template: 'address.html',
+      controller: require('./controllers/address'),
+      fields: [
+        'storage-address-manual'
+      ],
+      prereqs: ['/storage-postcode', '/storage-weapons-ammo', '/supporting-docs'],
```
joefitter: do we need this? I thought this was fixed

apps/new-dealer/index.js
```diff
@@ -103,12 +103,59 @@ module.exports = {
...
+      backLink: 'storage-postcode',
```
joefitter: and this

apps/new-dealer/index.js
bendiggle: @joefitter This is needed for the links that take you to the manual-address page. Without it, it takes you to the completely wrong page (even though the link is pointing the correct page)

apps/new-dealer/index.js
bendiggle: @joefitter This is needed as when you click the link that takes you to the manual address page, the back link is missing from the page. I assume because we are coming to the page via a link and not a page submission.

joefitter: ah fair do's. Is this what we do on GRO?

bendiggle: yep

PR #39, opened 2016-10-26
flho-178 authority-number-renew-vary step

bendiggle:
Added the authority-number-renew-vary step:
- Added the step to the routes config,
- Added the fields to the fields config,
- Added the custom html template,
- Added the translations for the page, fields and validation,
- Added the acceptance tests for the step and the page object,
- Changed the css class on the activity template and app.scss so it can be re-used for all summary-details sections

### DIFF
--- a/apps/new-dealer/acceptance/features/authority-number-renew-vary.js
+++ b/apps/new-dealer/acceptance/features/authority-number-renew-vary.js
@@ -1,0 +1,44 @@
+'use strict';
+
+const steps = require('../../');
+
+Feature('Authority number renew vary step');
+
+Before((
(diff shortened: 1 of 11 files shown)


apps/new-dealer/views/activity.html
```diff
@@ -4,7 +4,7 @@
       {{#renderField}}{{/renderField}}
     {{/fields}}
     <details>
-      <summary class="activity-details" role="button">
+      <summary class="summary-details" role="button">
```
JoeChapman: If this has a role of button, shouldn't the expanding area have a role?

PR #43, opened 2016-11-08
flho-163 storage-add-another-address

bendiggle:
This step is to allow the user to add multiple addresses whilst using the same steps to loop over.

What is basically happening is, when we get to the address-lookup step or manual-address step, we are adding the address to an array along with an id and then clearing the fields in order to go around again.

When the user has added an address, the address will appear on the add-another-address step where they are able to change the address, add more or continue.

If the user chooses to add another address, they are taken to the storage-postcode step where they can do the process again. The user will be able to see the addresses the have added just like the add-another-address step. See prototype -> https://s5-firearms-application.herokuapp.com/apply/storage-address-add

If the user chooses to change the address, then they are taken to the storage-postcode step with the fields entered with the relevant data. The url is appended with /edit/${id}. The id allows us to get the relevant data to fill the fields and allows us to overwrite the old address with the new one in the array.

When the user clicks back on the add-another-address step, then the user is taken back to the postcode-lookup with the url appended with /edit/${id}

As well as this, the usual step components have been added:
	* The steps have been added to the routes config,
	* The fields have been added to the fields config,
	* Controllers have been added for the steps,
	* The translations have been added for the page, fields and validation,
	* Custom html templates have been created,
	* Acceptance tests have been updated for storage-address including the page-object.

### DIFF
--- a/apps/new-dealer/controllers/address-lookup.js
+++ b/apps/new-dealer/controllers/address-lookup.js
@@ -22,7 +22,6 @@ module.exports = class AddressLookup extends BaseController {
       };
     });
 
-    // const field = this.options.locals.field;
     const count = `${formattedlist.length} addresses`;
     this.options.fields[`${field}-address-lookup`].options = [{value: count, label: count}].concat(formattedlist);
     super.getValues(req, res, callback);
--- a/apps/new-dealer/controllers/address-lookup.js
+++ b/apps/new-dealer/controllers/address-lookup.js
@@ -12,7 +12,8 @@ module.exports = class AddressLookup extends BaseController {
   }
 
   getValues(req, res, callback) {
-    const addresses = req.sessionModel.get('addresses');
+    const field = this.options.locals.field;
+    const addresses = req.sessionModel.get(`${field}-addresses`);
     const formattedlist = _.map(_.map(addresses, 'formatted_address'), address => {
       address = address.split('\n').join(', ');
       return {
@@ -21,7 +22,7 @@ module.exports = class AddressLookup extends BaseController {
       };
     });
 
-    const field = this.options.locals.field;
+    // const field = this.options.locals.field;
     const count = `${formattedlist.length} addresses`;
     this.options.fields[`${field}-address-lookup`].options = [{value: count, label: count}].concat(formattedlist);
     super.getValues(req, res, callback);
--- a/apps/new-dealer/controllers/postcode.js
+++ b/apps/new-dealer/controllers/postcode.js
@@ -8,6 +8,7 @@ const _ = require('lodash');
 
 module.exports = class PostcodeController extends BaseController {
   process(req, res, callback) {
+    const postcodesModel = new PostcodesModel();
     const field = this.options.locals.field;
     const postcode = req.form.values[`${field}-postcode`];
     const previousPostcode = req.sessionModel.get(`${field}-postcode`);
@@ -22,7 +23,6 @@ module.exports = class PostcodeController extends BaseController {
       return callback();
     }
 
-    const postcodesModel = new PostcodesModel();
     postcodesModel.fetch(postcode)
       .then(data => {
         if (data.length) {
--- a/apps/new-dealer/controllers/postcode.js
+++ b/apps/new-dealer/controllers/postcode.js
@@ -18,7 +18,7 @@ module.exports = class PostcodeController extends BaseController {
 
     if (_.startsWith(postcode, 'BT')) {
       req.sessionModel.unset('postcodeApiMeta');
-      req.sessionModel.unset('addresses');
+      req.sessionModel.unset(`${field}-addresses`);
       return callback();
     }
 
@@ -26,9 +26,9 @@ module.exports = class PostcodeController extends BaseController {
     postcodesModel.fetch(postcode)
       .then(data => {
         if (data.length) {
-          req.sessionModel.set('addresses', data);
+          req.sessionModel.set(`${field}-addresses`, data);
         } else {
-          req.sessionModel.unset('addresses');
+          req.sessionModel.unset(`${field}-addresses`);
           req.sessionModel.set('postcodeApiMeta', {
             messageKey: 'not-found'
           });
--- a/apps/new-dealer/controllers/storage-address-lookup.js
+++ b/apps/new-dealer/controllers/storage-address-lookup.js
@@ -1,0 +1,94 @@
+'use strict';
+const BaseController = require('hof').controllers.base;
+const ErrorController = require('hof').controllers.error;
+const _ = require('lodash');
+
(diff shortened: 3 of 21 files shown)


apps/new-dealer/controllers/address-lookup.js
```diff
@@ -21,7 +22,7 @@ module.exports = class AddressLookup extends BaseController {
       };
     });
 
-    const field = this.options.locals.field;
+    // const field = this.options.locals.field;
```
easternbloc: can this be deleted

apps/new-dealer/controllers/storage-address-lookup.js
```diff
@@ -0,0 +1,94 @@
+'use strict';
+const BaseController = require('hof').controllers.base;
+const ErrorController = require('hof').controllers.error;
+const _ = require('lodash');
+
+module.exports = class StorageAddressLookup extends BaseController {
+  locals(req, res) {
+    const locals = super.locals(req, res);
```
easternbloc: very minor gripe but maybe group const and let together when they're this close?

apps/new-dealer/controllers/storage-address.js
```diff
@@ -0,0 +1,82 @@
...
+          }));
+        }
+      }
+      return callback(null, values);
+    });
+  }
+
+  saveValues(req, res, callback) {
+    let storageAddresses = req.sessionModel.get('storageAddresses') || {};
+    const address = req.form.values['storage-address-manual'];
```
easternbloc: same here

apps/new-dealer/controllers/storage-postcode.js
```diff
@@ -0,0 +1,98 @@
...
+      return callback();
+    }
+
+    if (_.startsWith(postcode, 'BT')) {
+      req.sessionModel.unset('postcodeApiMeta');
+      req.sessionModel.unset('storage-addresses');
+      return callback();
+    }
+
+    const postcodesModel = new PostcodesModel();
```
easternbloc: define at the top of the function?

bendiggle:
@easternbloc Updated with your comments.

@gxxm @joefitter @JoeChapman

apps/new-dealer/controllers/storage-address.js
```diff
@@ -0,0 +1,82 @@
...
+        // eslint-disable-next-line no-console
+        console.error(err);
+        return callback(null, values);
+      }
+      if (req.params.action === 'manual') {
+            req.sessionModel.unset([
+              'storage-postcode',
+              'postcodeApiMeta'
+            ]);
+        if (req.params.id && req.params.id !== undefined) {
```
joefitter: do we need to check this twice?

apps/new-dealer/controllers/storage-address.js
```diff
@@ -0,0 +1,82 @@
...
+      return back;
+    }
+    return backLink;
+  }
+
+  getValues(req, res, callback) {
+    super.getValues(req, res, (err, values) => {
+      if (err) {
+        // eslint-disable-next-line no-console
+        console.error(err);
```
joefitter: shouldn't we be calling back and and passing the error here? is there a reason we are consoling it?

apps/new-dealer/views/storage-address.html
```diff
@@ -0,0 +1,27 @@
...
+        {{#t}}fields.storage-postcode.label{{/t}}
+      </label>
+      <p>
+        <span class="postcode">{{postcode}}</span>
+        <span class="link"><a href="{{baseUrl}}/storage-postcode">{{#t}}pages.address.edit{{/t}}</a></span>
+      </p>
+    {{/postcode}}
+    {{#postcodeApiMessageKey}}
+      <div class="alert-info">
+              <span class="info-message">
```
joefitter: indentation

bendiggle:
Updated @joefitter @easternbloc

apps/new-dealer/controllers/postcode.js
```diff
@@ -18,17 +18,17 @@ module.exports = class PostcodeController extends BaseController {
 
     if (_.startsWith(postcode, 'BT')) {
       req.sessionModel.unset('postcodeApiMeta');
-      req.sessionModel.unset('addresses');
+      req.sessionModel.unset(`${field}-addresses`);
       return callback();
     }
 
     const postcodesModel = new PostcodesModel();
```
easternbloc: this could probably go to the top of the function

bendiggle: I did it for storage-postcode.js but forgot to do it for the this. Will do it now

apps/new-dealer/controllers/storage-address.js
```diff
@@ -0,0 +1,82 @@
...
+    }
+    return backLink;
+  }
+
+  getValues(req, res, callback) {
+    super.getValues(req, res, (err, values) => {
+      if (err) {
+        // eslint-disable-next-line no-console
+        callback(err);
+        return callback(null, values);
```
joefitter: why are we calling callback twice? If there's an error we want to return callback(err) - dont worry about passing it values. Pretty sure this will actually break as headers are already set

apps/new-dealer/controllers/storage-postcode.js
```diff
@@ -0,0 +1,98 @@
...
+    }
+    return Object.assign({}, locals, {storageAddresses, hasStorageAddresses, id});
+  }
+
+  getValues(req, res, callback) {
+    super.getValues(req, res, (err, values) => {
+      if (err) {
+        // eslint-disable-next-line no-console
+        callback(err);
+        return callback(null, values);
```
joefitter: same here

apps/new-dealer/controllers/storage-postcode.js
```diff
@@ -0,0 +1,98 @@
...
+    if (req.params.action === 'edit') {
+      id = req.params.id;
+    }
+    return Object.assign({}, locals, {storageAddresses, hasStorageAddresses, id});
+  }
+
+  getValues(req, res, callback) {
+    super.getValues(req, res, (err, values) => {
+      if (err) {
+        // eslint-disable-next-line no-console
```
joefitter: dont need this

apps/new-dealer/controllers/storage-address.js
```diff
@@ -0,0 +1,82 @@
...
+      const back = req.params.action === 'edit' ? `${backLink}/${id}` : `${backLink}/edit/${id}`;
+      return back;
+    }
+    return backLink;
+  }
+
+  getValues(req, res, callback) {
+    super.getValues(req, res, (err, values) => {
+      if (err) {
+        // eslint-disable-next-line no-console
```
joefitter: dont need this

PR #56, opened 2016-11-15
flho-294 authority-number-renew-vary step changes. Changed the conten…

bendiggle:
…t on the step for the header and summary details. Swapped the customer number and authority number around


joefitter:
@bendiggle build failing

PR #62, opened 2016-11-23
Chore updated in line with so-master removal

joefitter:
* left pointing at bootstrap#development while in development
* installed hof-frontend-toolkit
* installed eslint-config-homeoffice
* updated so-acceptance to v4
* updated refs to hof

### DIFF
--- a/apps/new-dealer/models/postcodes.js
+++ b/apps/new-dealer/models/postcodes.js
@@ -1,8 +1,8 @@
 'use strict';
 
-const Model = require('hof').Model;
-const config = require('../../../config');
 const url = require('url');
+const Model = require('hmpo-model');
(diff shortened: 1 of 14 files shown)


apps/new-dealer/models/postcodes.js
```diff
@@ -1,8 +1,8 @@
 'use strict';
 
-const Model = require('hof').Model;
-const config = require('../../../config');
 const url = require('url');
+const Model = require('hmpo-model');
```
bendiggle: Will this eventually be hof-model? Or is there no reason to hof it up?

joefitter: one day dear Diggle, one day. But you are quite right - no reason to so far

PR #65, opened 2016-11-28
flho-175 - confirm page

bendiggle:
Added the confirm page to the form.
* Added confirm step to the routes config and sections to the steps,
* Added field to fields config and includeInSummary: false to some of the fields,
* Added a custom controller to manage the displaying of the storage address, the display of weapons and ammunition and it's quantities and the displaying of the contact holder sections,
* Added custom html for the step and html partials for the storageAddresses table and Weapons,ammuniton table,
* Added css for the page,
* Added tests for the step,
* Changed some of tests so that it test that it goes to the confirm page


apps/new-dealer/index.js
```diff
@@ -511,15 +547,23 @@ module.exports = {
...
     },
-    '/summary': {
-      clearSession: false
+    '/confirm': {
+      controller: require('./controllers/confirm'),
+      fields: [
+        'declaration'
+      ],
+      fieldsConfig: require('./fields'),
+      next: '/confirmation'
```
joefitter: you'll need email config here too, guess that will be a different ticket though?

PR #67, opened 2016-11-29
flho-313 Added the confirmation page to the form.

bendiggle:
* Added the step to the route config,
* Added custom html for page,
* Added translations for page,
* Added custom controller for translation logic for bullet points and for when the user selects arm-guards in the usage step,
* Added acceptance tests for the page.

### DIFF
--- a/apps/new-dealer/translations/src/en/pages.json
+++ b/apps/new-dealer/translations/src/en/pages.json
@@ -208,5 +208,36 @@
       "caseworker": "test",
       "customer": "test"
     }
+  },
+  "confirmation": {
+    "alert": "Application complete",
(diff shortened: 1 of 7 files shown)


apps/new-dealer/translations/src/en/pages.json
```diff
@@ -208,5 +208,36 @@
       "caseworker": "test",
       "customer": "test"
     }
+  },
+  "confirmation": {
+    "alert": "Application complete",
+    "email": "<p>We have sent an email to {{values.contact-email}} to confirm receipt of your application.</p><a href=#>Print or save the completed application</a>",
```
joefitter: todo?

easternbloc: Perhaps delete that based on our conversation?

PR #78, opened 2016-12-06
flho-338 Added the customer confirmation email

bendiggle:
* On the organisation step, changed the field other-specify to other-name as the email will show the organisation name,
* Changed the acceptance tests to reflect this change,
* In the confirm controller: 
                    - Moved the getting of the contact name into it's own function. 
                    - Added a getEmailConfig function. This gets some data (organisation name, contact name, reference number) and adds it to the config. 
                    - Adjust the intro text to greet the contact holder and use the contact's name.
* Change the confirm route config to send email to the contact's email address,
* Added the email text to the pages translations file,
* Created two email templates raw and formatted,
* Adjusted the config to send email,
* Added env variables to the docker-compose file

### DIFF
--- a/config.js
+++ b/config.js
@@ -14,12 +14,15 @@ module.exports = {
     }
   },
   email: {
-    caseworker: process.env.CASEWORKER_EMAIL || '',
     from: process.env.FROM_ADDRESS || '',
     replyTo: process.env.REPLY_TO || '',
     accessKeyId: process.env.AWS_USER || '',
     secretAccessKey: process.env.AWS_PASSWORD || '',
-    transportType: 'ses',
-    region: process.env.EMAIL_REGION || ''
+    transportType: process.env.EMAIL_TRANSPORT,
+    region: process.env.EMAIL_REGION || '',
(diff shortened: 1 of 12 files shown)


config.js
```diff
@@ -14,12 +14,15 @@ module.exports = {
...
-    caseworker: process.env.CASEWORKER_EMAIL || '',
     from: process.env.FROM_ADDRESS || '',
     replyTo: process.env.REPLY_TO || '',
     accessKeyId: process.env.AWS_USER || '',
     secretAccessKey: process.env.AWS_PASSWORD || '',
-    transportType: 'ses',
-    region: process.env.EMAIL_REGION || ''
+    transportType: process.env.EMAIL_TRANSPORT,
+    region: process.env.EMAIL_REGION || '',
+    ignoreTLS: process.env.IGNORE_TLS === 'true',
```
easternbloc: ;) thinking we should probably make hof-emailer a bit more defensive around this but good for now

easternbloc:
@joefitter these changes (making compose work) needs to be reflected into end-tenancy

PR #84, opened 2016-12-09
flho-344 club secretary address step.

bendiggle:
* Added the step to the routes config. 
* Added the fields to the fields config. 
* Added the translations for pages, fields and validation. 
* Added acceptance tests for the step


sulthan-ahmed:
looks good @bendiggle just a question, all the functionality isn’t there right? as in the app couldn’t find some real addresses e.g. n22

PR #85, opened 2016-12-09
Feature/flho 347 second contact address

bendiggle:
* Added then step to the routes config. 
* Added the fields to the fields config. 
* Added the translations for the page, fields and validation. 
* Added the acceptance tests for the steps.

### DIFF
--- a/apps/shooting-clubs/fields/index.js
+++ b/apps/shooting-clubs/fields/index.js
@@ -60,5 +60,25 @@ module.exports = {
   'club-secretary-phone': {
     mixin: 'input-text',
     validate: 'required'
+  },
+  'second-contact-postcode': {
+    mixin: 'input-text-code',
+    validate: ['required', 'postcode'],
+    formatter: 'uppercase',
+    includeInSummary: false
+  },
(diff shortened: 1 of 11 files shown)


apps/shooting-clubs/fields/index.js
```diff
@@ -52,5 +52,25 @@ module.exports = {
...
+    className: 'address',
+    includeInSummary: false
+  },
+  'second-contact-address-manual': {
+    mixin: 'textarea',
+    validate: 'required',
+    'ignore-defaults': true,
+    formatter: ['trim', 'hyphens'],
+    attributes: [{
+      attribute: 'height',
```
joefitter: this isn't a thing. `<textarea rows="5" cols="10"/>`

bendiggle: @joefitter Could have said something about 2 months ago! So straight on the html template, no attributes here?

bendiggle: @joefitter Never mind. Done it

bendiggle:
I have updated all textareas @joefitter

PR #88, opened 2016-12-12
Refactored storage-address loop so it can be reused

bendiggle:
Refactored the address looping so that it can be used for more than one step. 
* Refactored the storage-postcode, storage-address, storage-address-lookup and storage-add-another-address controllers. 
* Renamed the controllers and moved them into common. 
* Changed the references of them in the routes config. 
* Moved the views into common.
* Renamed the views and referenced them in the routes config. 
* Changed the variable names that are passed to the views via locals.


apps/common/controllers/add-another-address-loop.js
```diff
@@ -0,0 +1,48 @@
+'use strict';
+
+const _ = require('lodash');
+const BaseController = require('./base');
+
+module.exports = class AddAnotherAddressController extends BaseController {
+  locals(req, res, callback) {
```
joefitter: locals doesn't accept callback

apps/common/controllers/add-another-address-loop.js
```diff
@@ -0,0 +1,48 @@
...
+    const locals = super.locals(req, res, callback);
```
joefitter: as above

apps/common/controllers/add-another-address-loop.js
```diff
@@ -0,0 +1,48 @@
...
+    const field = locals.field;
```
joefitter: what happens if step.locals.field is undefined? We should probably throw an informative error here to help out future-us :)

apps/common/controllers/add-another-address-loop.js
```diff
@@ -0,0 +1,48 @@
...
+    const addresses = req.sessionModel.get(`${field}Addresses`);
+    const hasAddresses = req.sessionModel.get(`${field}Addresses`) ? true : false;
+    const items = [];
+    _.forEach(addresses, (value, key) => {
+      const address = {
+        id: key,
+        address: value.address
+      };
+      items.push(address);
+    });
```
joefitter: Firstly, we use `_.each` in quite a few places in the codebase - I would prefer us to keep using `_.each` rather than an alias.

Secondly - would it not be better to create the items array in one operation?
```js
const items = Object.keys(addresses).map(key => ({
  id: key,
  address: addresses[key].address
});

or

const items = _.map(addresses, (value, key) => ({
  id: key,
  address: value.address
}));
```

apps/common/controllers/add-another-address-loop.js
```diff
@@ -0,0 +1,48 @@
...
+    return Object.assign({}, locals, {
+      items,
+      hasAddresses
+    });
+  }
+
+  get(req, res, callback) {
+    if (req.params.action === 'delete' && req.params.id) {
+      const field = this.options.locals.field;
```
joefitter: if we are relying on field being set throughout this controller we should probably set a default in the constructor, or throw an error

apps/common/controllers/add-another-address-loop.js
```diff
@@ -0,0 +1,48 @@
...
+  removeItem(req, res, field) {
+    const items = req.sessionModel.get(`${field}Addresses`);
+    req.sessionModel.set(`${field}Addresses`, _.omit(items, req.params.id));
+    const step = _.size(items) > 1 ? `/${field}-add-another-address` : `/${field}-postcode`;
+    return res.redirect(`${req.baseUrl}${step}`);
+  }
+
+  getBackLink(req) {
+    const field = this.options.locals.field;
+    const app = this.options.locals.app;
```
joefitter: app?

joefitter: can we not find this from req.app.path()?

apps/common/controllers/add-another-address-loop.js
```diff
@@ -0,0 +1,48 @@
...
+    const addresses = req.sessionModel.get(`${field}Addresses`);
+    const id = _.last(Object.keys(addresses));
+    return `/${app}/${field}-postcode/edit/${id}`;
```
joefitter: if app isn't defined you will get a broken `//` at the start of your url

apps/common/controllers/add-another-address-loop.js
```diff
@@ -0,0 +1,51 @@
...
+const BaseController = require('./base');
+
+module.exports = class AddAnotherAddressController extends BaseController {
+  constructor(options) {
+    super(options);
+    if (options.locals.field) {
+      this.field = options.locals.field;
+    } else {
+      // eslint-disable-next-line no-console
+      console.error('No field set in locals');
```
joefitter: what happens if options.locals.field isn't set? Can it be omitted? if not then this should actually throw an error

apps/common/controllers/add-another-address-loop.js
```diff
@@ -0,0 +1,51 @@
...
+    }));
+    return Object.assign({}, locals, {
+      items,
+      hasAddresses
+    });
+  }
+
+  get(req, res, callback) {
+    if (req.params.action === 'delete' && req.params.id) {
+      return this.removeItem(req, res, this.field);
```
joefitter: we don't need to pass this.field as the remove method can now access it from the instance

apps/common/controllers/address-lookup-loop.js
```diff
@@ -0,0 +1,121 @@
...
+const BaseController = require('./base');
+
+module.exports = class AddressLookupLoopController extends BaseController {
+  constructor(options) {
+    super(options);
+    if (options.locals.field) {
+      this.field = options.locals.field;
+    } else {
+      // eslint-disable-next-line no-console
+      console.error('No field set in locals');
```
joefitter: as above

joefitter: do we not want to set a default? this.field = options.locals.field || 'blah'

apps/common/controllers/add-another-address-loop.js
bendiggle: @joefitter It can't be omitted and I don't think a default value can be set. Is my console.error not acceptable?

joefitter: It would be better IMO - if field is _required_ for the controller to work - throwing a big fat error would be preferable to a piddly console statement @JoeChapman @easternbloc what say you?

PR #91, opened 2016-12-15
flho-352 and flho-353. Storage address loop to the shootin clubs

bendiggle:
- Added the steps to the routes config (postcode, address, address-lookup and add-another-address),
- Added the fields to the fields config,
- Added the translations,
- Added the acceptance tests for the steps,
- Moved buttons translations to common.


apps/shooting-clubs/acceptance/features/storage-address.js
```diff
@@ -0,0 +1,187 @@
...
+  storageAddressPage
+) => {
+  I.visitPage(storageAddressPage, steps);
+});
+
+Scenario('The correct form elements are present on storage-postcode step', (
+  I,
+  storageAddressPage
+) => {
+  I.seeElements([
```
joefitter: this accepts a string or an array, I'd just pass the single item if there's only one. Or you could use the build in method seeElement - up to you!

apps/shooting-clubs/acceptance/features/storage-address.js
```diff
@@ -0,0 +1,187 @@
...
+  I.click(storageAddressPage.links['manual-entry']);
+  I.seeInCurrentUrl(storageAddressPage['address-url']);
+});
+
+Scenario('The correct form elements are present for storage-manual address step', (
+  I,
+  storageAddressPage
+) => {
+  I.click(storageAddressPage.links['manual-entry']);
+  I.seeElements([
```
joefitter: as above

apps/shooting-clubs/acceptance/features/storage-address.js
```diff
@@ -0,0 +1,187 @@
...
+Scenario('When I select No on add-another-address I am taken to the confirm page', (
+  I,
+  storageAddressPage,
+  confirmPage
+) => {
+  storageAddressPage.selectAddressAndSubmit();
+  I.click(storageAddressPage.fields.no);
+  I.submitForm();
+  I.seeInCurrentUrl(confirmPage.url);
+});
```
joefitter: How come the tests for this functionality are here? Shouldn't it be in common if it is shared across journeys now? We don't need to test the same functionality twice, just that it goes to the right place

bendiggle: @joefitter You're right. This should go into common! I will do that friend

bendiggle: @joefitter This might be a bit more difficult than first thought as this is the address process as a whole rather than individual steps

joefitter:
@sulthan-ahmed I stand by what I said about moving this stuff to common, but yeah lets catch up in the new year

PR #97, opened 2017-01-09
Feature/flho 350 shooting club categories

sulthan-ahmed:
* Add the shooting club categories page
* Remove duplicate pages of storage address
* Amend the tests for location address page to work with shooting club categories page
* Added unit tests 

Story https://jira.digital.homeoffice.gov.uk/browse/FLHO-350
Prototype example https://shooting-club.herokuapp.com/location/location-address-category


apps/shooting-clubs/acceptance/features/location-address.js
```diff
@@ -134,11 +134,66 @@ Scenario('from manual entry step: I am taken to the location-add-another-address
...
+Scenario('category step: when I select 2 categories: full-bore rifles, small-bore rifles category, I see these categories against the address', (
+  I,
+  locationAddressPage
+) => {
+  locationAddressPage.selectAddressAndSubmit();
+  I.click(locationAddressPage.category.field.fullBoreRifles);
+  I.click(locationAddressPage.category.field.smallBoreRifles);
+  I.submitForm();
+  I.see(locationAddressPage.category.content.fullBoreRifles, locationAddressPage.addAnotherAddress.field.summary)
+  I.see(locationAddressPage.category.content.smallBoreRifles, locationAddressPage.addAnotherAddress.field.summary)
```
sulthan-ahmed: I could use .seeEach with an array but this does not take a selector.  It just searches the whole page for text which can cause failures of tests if texts are duplicated.  It is also a major reason for slowing down acceptance tests for the registered traveller project in the past.

sulthan-ahmed: maybe I could create a function that looks for multiple items in an array

apps/shooting-clubs/acceptance/features/location-address.js
```diff
@@ -134,11 +134,66 @@ Scenario('from manual entry step: I am taken to the location-add-another-address
...
   locationAddressPage
 ) => {
   locationAddressPage.selectAddressAndSubmit();
+  I.click(locationAddressPage.category.field.fullBoreRifles);
+  I.click(locationAddressPage.category.field.smallBoreRifles);
+  I.click(locationAddressPage.category.field.muzzleLoadingPistols);
+  I.submitForm();
+  I.see(locationAddressPage.category.content.fullBoreRifles, locationAddressPage.addAnotherAddress.field.summary)
+  I.see(locationAddressPage.category.content.smallBoreRifles, locationAddressPage.addAnotherAddress.field.summary)
+  I.see(locationAddressPage.category.content.muzzleLoadingPistols, locationAddressPage.addAnotherAddress.field.summary)
```
sulthan-ahmed: As above

apps/shooting-clubs/views/location-address-category.html
```diff
@@ -0,0 +1,9 @@
+{{<partials-page}}
+  {{$page-content}}
+    {{#fields}}
+      {{#renderField}}{{/renderField}}
+    {{/fields}}
+    {{#input-submit}}continue{{/input-submit}}
+    {{> partials-add-another-address-summary}}
+  {{/page-content}}
+{{/partials-page}}
```
lennym: This looks *very* similar to https://github.com/UKHomeOffice/firearms/blob/master/apps/common/views/add-another-address-loop.html - it's probably better to re-use that than create a new template.

sulthan-ahmed: Good spot, I'll give that a try

apps/common/controllers/add-another-address-loop.js
```diff
@@ -19,14 +19,28 @@ module.exports = class AddAnotherAddressController extends BaseController {
...
-      hasAddresses
+      hasAddresses,
+      hasCategories: _.sample(addresses).categories !== undefined
     });
   }
 
+  translateCategories(req, values) {
+    if (!Array.isArray(values)) {
+      values = [values];
+    }
```
JoeChapman: https://lodash.com/docs/4.17.4#castArray

apps/common/controllers/add-another-address-loop.js
```diff
@@ -19,14 +19,28 @@ module.exports = class AddAnotherAddressController extends BaseController {
...
+    const categories = values.map(value => {
+      let key = `fields.location-address-category.options.${value}.label`;
+      let result = req.translate(key);
+      return result === key ? value : result;
+    }).join('\n');
+    return categories;
```
JoeChapman: If `categories` is only used once, you could simply return the result of `values.map`

apps/common/controllers/address-lookup-loop.js
```diff
@@ -34,10 +36,23 @@ module.exports = class AddressLookupLoopController extends BaseController {
...
-      id
+      id,
+      hasCategories
     });
   }
 
+  translateCategories(req, values) {
+    if (!Array.isArray(values)) {
+      values = [values];
+    }
```
JoeChapman: https://lodash.com/docs/4.17.4#castArray

apps/common/controllers/address-lookup-loop.js
```diff
@@ -34,10 +36,23 @@ module.exports = class AddressLookupLoopController extends BaseController {
...
+    const categories = values.map(value => {
+      let key = `fields.location-address-category.options.${value}.label`;
+      let result = req.translate(key);
+      return result === key ? value : result;
+    }).join('\n');
+    return categories;
+  }
```
JoeChapman: I'm in danger of repeating myself but, if `categories` is only used once, you could simply return the result of `values.map`

There is quite a lot of duplication of code between this and apps/common/controllers/add-another-address-loop.js

apps/common/controllers/address-loop.js
```diff
@@ -19,18 +19,33 @@ module.exports = class AddressLoopController extends BaseController {
...
+      hasCategories,
       postcode,
       postcodeApiMessageKey: isManual ? '' : (req.sessionModel.get('postcodeApiMeta') || {}).messageKey
     });
   }
 
+  translateCategories(req, values) {
+    if (!Array.isArray(values)) {
+      values = [values];
+    }
```
JoeChapman: https://lodash.com/docs/4.17.4#castArray

apps/common/controllers/postcode-loop.js
```diff
@@ -18,15 +18,34 @@ module.exports = class PostcodeLoopController extends BaseController {
...
+      hasAddresses,
+      id,
+      hasCategories
+    });
+  }
+
+  translateCategories(req, values) {
+    if (!Array.isArray(values)) {
+      values = [values];
+    }
```
JoeChapman: https://lodash.com/docs/4.17.4#castArray

apps/common/controllers/add-another-address-loop.js
joefitter: castArray does exactly the same as these 3 lines of code, however for the sake of duplication it would be better IMO to use the lodash method

sulthan-ahmed: I did not know that, niiiiiice

apps/common/controllers/address-lookup-loop.js
sulthan-ahmed: say that again?

apps/common/controllers/address-loop.js
sulthan-ahmed: o that

apps/common/controllers/base-address.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+const BaseController = require('./base');
+const _ = require('lodash');
+
+module.exports = class BaseAddressController extends BaseController {
+	translateCategories(req, values) {
+    return _.castArray(values).map(value => {
+      let key = `fields.location-address-category.options.${value}.label`;
+      let result = req.translate(key);
```
JoeChapman: `key` and `result` should be immutable `const` vars

sulthan-ahmed: 👍

apps/common/controllers/add-another-address-loop.js
```diff
@@ -17,13 +17,11 @@ module.exports = class AddAnotherAddressController extends BaseController {
...
-      id: key,
-      address: value.address
-    }));
+    const items = this.mapAddress(addresses, req);
     return Object.assign({}, locals, {
       items,
-      hasAddresses
+      hasAddresses,
+      hasCategories: this.hasCategories(hasAddresses, addresses)
     });
```
JoeChapman: Better thanks @sulthan-ahmed. Are you going to move this into the parent class too?

apps/common/controllers/add-another-address-loop.js
```diff
@@ -17,13 +17,11 @@ module.exports = class AddAnotherAddressController extends BaseController {
...
-    const items = _.map(addresses, (value, key) => ({
-      id: key,
-      address: value.address
-    }));
+    const items = this.mapAddress(addresses, req);
     return Object.assign({}, locals, {
       items,
-      hasAddresses
+      hasAddresses,
+      hasCategories: this.hasCategories(hasAddresses, addresses)
```
JoeChapman: Please make this line consistent with `hasAddreses`

apps/common/controllers/add-another-address-loop.js
```diff
@@ -17,13 +17,11 @@ module.exports = class AddAnotherAddressController extends BaseController {
     const locals = super.locals(req, res);
     const addresses = req.sessionModel.get(`${this.field}Addresses`);
     const hasAddresses = req.sessionModel.get(`${this.field}Addresses`) ? true : false;
-    const items = _.map(addresses, (value, key) => ({
-      id: key,
-      address: value.address
-    }));
+    const items = this.mapAddress(addresses, req);
```
JoeChapman: `mapAddresses`

apps/common/controllers/add-another-address-loop.js
joefitter: @sulthan-ahmed I thought we did move this - must have been an oversight!

apps/common/controllers/postcode-loop.js
```diff
@@ -35,9 +18,9 @@ module.exports = class PostcodeLoopController extends BaseController {
         return callback(err);
       }
       if (req.params.action === 'edit') {
-        const address = values[`${this.field}Addresses`][req.params.id];
+        const address = values[`${this.options.locals.field}Addresses`][req.params.id];
```
joefitter: `this.options.addressKey`

apps/common/controllers/postcode-loop.js
```diff
@@ -50,7 +33,8 @@ module.exports = class PostcodeLoopController extends BaseController {
 
   saveValues(req, res, cb) {
     const saveValues = super.saveValues(req, res, cb);
-    this.postcodeChanged = req.form.values[`${this.field}-postcode`] !== req.sessionModel.get(`${this.field}-postcode`);
+    this.postcodeChanged = req.form.values[`${this.options.locals.field}-postcode`]
```
joefitter: don't use the instance - save to sessionModel

apps/common/controllers/postcode-loop.js
```diff
@@ -70,33 +54,35 @@ module.exports = class PostcodeLoopController extends BaseController {
   }
 
   removeItem(req, res) {
-    const items = req.sessionModel.get(`${this.field}Addresses`);
-    req.sessionModel.set(`${this.field}Addresses`, _.omit(items, req.params.id));
-    const step = _.size(items) > 1 ? `/${this.field}-add-another-address` : `/${this.field}-postcode`;
+    const items = req.sessionModel.get(`${this.options.locals.field}Addresses`);
```
joefitter: `this.options.addressKey`

apps/common/controllers/postcode-loop.js
```diff
@@ -70,33 +54,35 @@ module.exports = class PostcodeLoopController extends BaseController {
...
+    req.sessionModel.set(`${this.options.locals.field}Addresses`,
```
joefitter: `this.options.addressKey`

apps/common/controllers/postcode-loop.js
joefitter: this removing address functionality could be abstracted to the baseAddress controller as its used by every page

test/common/controllers/base-address.js
```diff
@@ -0,0 +1,154 @@
...
+
+StubBaseController.prototype.locals = sinon.stub();
+
+const BaseAddressController = proxyquire('../../../apps/common/controllers/base-address', {
+  './base': StubBaseController
+});
+
+describe('BaseAddressController', function() {
+
+  it('extends base-controller', ()=> {
```
joefitter: space between `() => {` lint should pick this up

test/common/controllers/base-address.js
```diff
@@ -0,0 +1,154 @@
...
+    BaseAddressController.prototype.should.be.an.instanceOf(StubBaseController);
+  });
+
+  it('has a locals, translateCategories, mapAddress & hasCategories methods', ()=> {
```
joefitter: space

test/common/controllers/base-address.js
```diff
@@ -0,0 +1,154 @@
...
+const proxyquire = require('proxyquire');
+class StubBaseController {}
+
+StubBaseController.prototype.locals = sinon.stub();
+
+const BaseAddressController = proxyquire('../../../apps/common/controllers/base-address', {
+  './base': StubBaseController
+});
+
+describe('BaseAddressController', function() {
```
joefitter: did you mean to use `function` here?

test/common/controllers/base-address.js
```diff
@@ -0,0 +1,154 @@
...
+
+  it('extends base-controller', ()=> {
+    BaseAddressController.prototype.should.be.an.instanceOf(StubBaseController);
+  });
+
+  it('has a locals, translateCategories, mapAddress & hasCategories methods', ()=> {
+    BaseAddressController.prototype.should.have.property('locals').that.is.a('function');
```
joefitter: can we add separate tests for these please? Ideally one assertion per test

apps/common/controllers/add-another-address-loop.js
```diff
@@ -35,16 +12,17 @@ module.exports = class AddAnotherAddressController extends BaseController {
   }
 
   removeItem(req, res) {
-    const items = req.sessionModel.get(`${this.field}Addresses`);
-    req.sessionModel.set(`${this.field}Addresses`, _.omit(items, req.params.id));
-    const step = _.size(items) > 1 ? `/${this.field}-add-another-address` : `/${this.field}-postcode`;
+    const items = req.sessionModel.get(`${this.options.locals.field}Addresses`);
+    req.sessionModel.set(`${this.options.locals.field}Addresses`, _.omit(items, req.params.id));
```
JoeChapman: This `${this.options.locals.field}Addresses` is fugly, and done twice in this method, and at least one other method. And `${this.options.locals.field}-postcode`. 
My feeling is that you should have a helper method that creates these strings.

test/common/controllers/base-address.js
```diff
@@ -0,0 +1,154 @@
+'use strict';
+
+const proxyquire = require('proxyquire');
+class StubBaseController {}
+
+StubBaseController.prototype.locals = sinon.stub();
```
joefitter: how come you are stubbing this here, then again below?

test/common/controllers/base-address.js
```diff
@@ -0,0 +1,154 @@
...
+  describe('instance', ()=> {
+    let baseAddressController;
+    const options = {
+      addressKey: 'foo'
+    };
+
+    beforeEach(()=> {
+      StubBaseController.prototype.locals = sinon.stub();
+      baseAddressController = new BaseAddressController(options);
+      baseAddressController.options = options;
```
joefitter: you could add a constructor to the stub controller to do this - 

```js
class StubBaseController {
  constructor(options) {
    this.options = options
  }
}
```

test/common/controllers/base-address.js
```diff
@@ -0,0 +1,154 @@
...
+          };
+          baseAddressController.locals(req, res).should.deep.equal(expectedVal);
+        });
+
+        it('calls base locals', () => {
+          StubBaseController.prototype.locals.should.not.have.been.called;
+
+          baseAddressController.locals(req, res);
+
+          StubBaseController.prototype.locals.should.have.been.called;
```
joefitter: I prefer to be explicit with the expected amount of calls - use calledOnce, calledTwice etc

apps/common/controllers/add-another-address-loop.js
```diff
@@ -35,16 +12,17 @@ module.exports = class AddAnotherAddressController extends BaseController {
...
+    const step = _.size(items) > 1 ?
+      `/${this.options.locals.field}-add-another-address` : `/${this.options.locals.field}-postcode`;
     return res.redirect(`${req.baseUrl}${step}`);
   }
```
JoeChapman: This method is doing more than just removing an item. 
I would consider splitting out the code that determines which step to redirect to.

test/common/controllers/base-address.js
```diff
@@ -0,0 +1,154 @@
...
+        it('returns an expected address from the sessionModel', ()=> {
+          baseAddressController.getAddresses(req).should.equal(fooValue);
+        });
+
+        it('returns undefined if the options cannot be found in the sessionModel', () => {
+          const otherOptions = {
+            addressKey: 'blah'
+          };
+          baseAddressController.options = otherOptions;
+          should.equal(baseAddressController.getAddresses(req), undefined);
```
joefitter: you could also use `expect` for this:

```js
expect(value).to.be.undefined;
```

currently you would need to use `chai.expect` but after this is rebased with Lenny's PR expect will be global

test/common/controllers/base-address.js
```diff
@@ -0,0 +1,154 @@
...
+          const addresses = [{address: 'London', postcode: '', categories: ['a', 'b']},
+            {address: 'Paris', postcode: '', categories: ['a', 'b', 'c']}];
+          const expected = [{id: 0, address: 'London', categories: 'a\nb'},
+            {id: 1, address: 'Paris', categories: 'a\nb\nc'}];
+          const act = baseAddressController.mapAddress(addresses, req);
+          act.should.deep.equal(expected);
+        });
+      });
+  });
+});
```
joefitter: this spacing/tabbing in this file looks pretty borked!

apps/common/controllers/add-another-address-loop.js
joefitter:
```js
`${this.options.locals.field}Addresses` === this.options.addressKey
```

apps/common/controllers/add-another-address-loop.js
joefitter: yeah, I touched on this in my review - removing an item is shared functionality on every page so should be abstracted to the base address controller

test/common/controllers/base-address.js
sulthan-ahmed: lint doesn't but I will change

test/common/controllers/base-address.js
lennym: I disagree fundamentally with "Ideally one assertion per test".

Why?

apps/common/controllers/base-address.js
```diff
@@ -0,0 +1,48 @@
...
+
+  locals(req, res) {
+    const locals = super.locals(req, res);
+    let addresses = this.getAddresses(req);
+    const hasAddresses = !!_.size(addresses);
+    const hasCategories = this.hasCategories(hasAddresses, addresses);
+    if (this.options.addressKey === 'locationAddresses') {
+      addresses = _.filter(addresses, value => value.categories !== undefined);
+    }
+    const items = this.mapAddress(addresses, req);
```
JoeChapman: `mapAddresses`?

test/common/controllers/base-address.js
lennym: I'd just add a `global.expect = chai.expect;` line to `test/common.js` and deal with the conflict later.

apps/common/controllers/base-address.js
```diff
@@ -0,0 +1,48 @@
...
+    return Object.assign({}, locals, {
+      items,
+      hasAddresses,
+      hasCategories
+    });
+  }
+
+  getAddresses(req) {
+    return req.sessionModel.get(this.options.addressKey);
+  }
```
JoeChapman: I think @lennym was correct, this method is unnecessary. The return value from `req.sessionModel.get` can be implicitly tested from `locals()`

apps/common/controllers/base-address.js
```diff
@@ -0,0 +1,48 @@
...
+
+  translateCategories(req, values) {
+    return _.castArray(values).map(value => {
+      const key = `fields.location-address-category.options.${value}.label`;
+      const result = req.translate(key);
+      return result === key ? value : result;
+    }).join('\n');
+  }
+
+  mapAddress(addresses, req) {
```
JoeChapman: `mapAddresses`?

apps/common/controllers/base-address.js
```diff
@@ -0,0 +1,48 @@
...
+    return _.map(addresses, (value, key) => ({
+      id: key,
+      address: value.address,
+      categories: this.translateCategories(req, value.categories)
+    }));
+  }
+
+  hasCategories(hasAddresses, addresses) {
+    return !!(hasAddresses ? _.sample(addresses).categories !== undefined : false);
+  }
```
JoeChapman: some lines of code in this PR aren't immediately clear.
This line seems confused, it casts `false` to `false` if `hasAddresses` is not truthy.

apps/common/controllers/postcode-loop.js
```diff
@@ -70,33 +54,35 @@ module.exports = class PostcodeLoopController extends BaseController {
...
+      _.omit(items, req.params.id));
+    const step = _.size(items) > 1 ? `/${this.options.locals.field}-add-another-address` :
+      `/${this.options.locals.field}-postcode`;
```
JoeChapman: Again, this code is not responsible for removing an item so should not be in this method.
Also, it happens more than once, could be pushed upstream to parent class.

apps/common/controllers/base-address.js
lennym: This particular function is totally mad and should be refactored with extreme prejudice.

apps/common/controllers/postcode-loop.js
```diff
@@ -70,33 +54,35 @@ module.exports = class PostcodeLoopController extends BaseController {
...
   }
 
   process(req, res, callback) {
     const postcodesModel = new PostcodesModel();
-    const postcode = req.form.values[`${this.field}-postcode`];
-    const previousPostcode = req.sessionModel.get(`${this.field}-postcode`);
+    const postcode = req.form.values[`${this.options.locals.field}-postcode`];
+    const previousPostcode = req.sessionModel.get(`${this.options.locals.field}-postcode`);
     if (!postcode
       || previousPostcode && previousPostcode === postcode) {
```
JoeChapman: ?

apps/common/controllers/add-another-address-loop.js
```diff
@@ -35,16 +12,17 @@ module.exports = class AddAnotherAddressController extends BaseController {
...
 
   removeItem(req, res) {
-    const items = req.sessionModel.get(`${this.field}Addresses`);
-    req.sessionModel.set(`${this.field}Addresses`, _.omit(items, req.params.id));
-    const step = _.size(items) > 1 ? `/${this.field}-add-another-address` : `/${this.field}-postcode`;
+    const items = req.sessionModel.get(`${this.options.locals.field}Addresses`);
```
sulthan-ahmed: 👍

test/common/controllers/base-address.js
joefitter: we need to fix lint :(

test/common/controllers/base-address.js
joefitter: because if one assertion fails, then the test fails, but you don't immediately know which assertion from the test wording

apps/common/controllers/base-address.js
JoeChapman: I can't compete with @lennym's skill of clear and expressive speech

test/common/controllers/base-address.js
JoeChapman: And assert that it is called with `options`

test/common/controllers/base-address.js
JoeChapman: It should depend on how broad the description is.

joefitter: `it('tests all the functionality', () => {})`

joefitter: I don't feel particularly strongly about it, but its just something I have always done

lennym: "one assertion per test" is IMO entirely analogous to "one line of code per function". 

Use as many assertions as are needed to completely test the functionality described by the test. Certainly having *loads* of assertions is severe code smell, and can indicate the scope of the test is too broad, but if the test is readable and understandable (like this one) then it's totally fine.

apps/common/controllers/address-lookup-loop.js
```diff
@@ -47,12 +27,12 @@ module.exports = class AddressLookupLoopController extends BaseController {
     if (req.params.action === 'edit') {
       const steps = req.sessionModel.get('steps');
       _.remove(steps, step => {
```
sulthan-ahmed: ?? can we chat about this one when you're free @joefitter ?

apps/common/controllers/address-lookup-loop.js
```diff
@@ -62,7 +42,8 @@ module.exports = class AddressLookupLoopController extends BaseController {
     });
 
     const count = `${formattedlist.length} addresses`;
-    this.options.fields[`${this.field}-address-lookup`].options = [{value: count, label: count}].concat(formattedlist);
+    this.options.fields[`${this.options.locals.field}-address-lookup`].options =
+      [{value: count, label: count}].concat(formattedlist);
```
sulthan-ahmed: ?? and this one?

apps/common/controllers/address-lookup-loop.js
lennym: I agree as a general programming principle, but I'd leave it for another day in the interests of getting this in. It works as it is

apps/common/controllers/address-lookup-loop.js
lennym: This is going to need some changes to hof-template-mixins to be able to set dynamic options on a select/radio group. I will be picking that work up.

apps/common/controllers/address-lookup-loop.js
```diff
@@ -109,9 +92,9 @@ module.exports = class AddressLookupLoopController extends BaseController {
   }
 
   validateField(key, req) {
-    if (req.form.values[key] === this.options.fields[`${this.field}-address-lookup`].options[0].value) {
-      return new ErrorController(`${this.field}-address-lookup`, {
-        key: `${this.field}-address-lookup`,
+    if (req.form.values[key] === this.options.fields[`${this.options.locals.field}-address-lookup`].options[0].value) {
```
lennym: A I right in thinking this checks to see if the submitted value matches the first "null" option from the select ("please select" or similar) and throws if it matches?

apps/common/controllers/postcode-loop.js
lennym: @sulthan-ahmed To fix this, look for where the value of `this.postcodeChanged` is used, then swap `this.postcodeChanged` for `req.sessionModel.get('postcode-changed')` and `req.sessionModel.set('postcodechanged', ...)` here.

Then the `postcodeChanged` property is specific to the current user, and not shared across all implementations of the controller.

sulthan-ahmed:
@lennym there seems to be a bug when removing addresses.  If you attempt to remove all the addresses than you can except the last one.  Then when you add more addresses, it won't let you delete them.  One to investigate after I've done my other work for this

apps/common/controllers/base-address.js
```diff
@@ -0,0 +1,50 @@
...
+  hasCategories(hasAddresses, addresses) {
+    if (hasAddresses) {
+      const firstAddressItem = addresses[Object.keys(addresses)[0]];
+      return _.has(firstAddressItem, 'categories');
+    }
+    else {
+      return false;
+    }
+  }
+};
```
sulthan-ahmed: to understand this a bit further, addresses is made up like so
addresses>> { '2':
   { address: 'Paris',
     postcode: '',
     categories: [ 'small-bore-rifles', 'muzzle-loading-pistols' ] },
  '3':
   { address: 'Jordan',
     postcode: '',
     categories:
      [ 'full-bore-rifles',
        'small-bore-rifles',
        'muzzle-loading-pistols' ] } }
hasCategories just checks that any child object has a property called categories.  I've attempted to make it more readable and not use the _.sample function

Let me know if it can be improved

lennym:
I've removed all the option mutation from this PR, and got the acceptance tests passing, so theoretically this should be good to merge now.

I am curious to know what's happened to the storage address steps though. They've been *mostly* deleted from the steps config, and were half deleted from the acceptance tests when I picked this up, so I'm a bit confused.

lennym:
Yeah, it went *waaay* out of scope. I think we need to learn a lesson from this. Not sure what that lesson is.

PR #100, opened 2017-01-11
Supporting document upload

lennym:
Adds two steps to the end of the form, asking if you want to upload documents, and then doing that.

Documents are sent to a downstream API (currently mocked) and then the returned URLs are added to an array `supporting-documents` on the session.

The prompt/upload loop is continued until the user says "No" to uploading any further documents, at which point they are taken to the "Confirm" page. 

It is not necessary to upload any documents at all.


lennym:
Now with 100% unit test coverage for the new components.

🎉 🎉 :tada:

apps/common/models/file-upload.js
```diff
@@ -0,0 +1,33 @@
+'use strict';
+
+const url = require('url');
+
+const Model = require('hmpo-model');
```
JoeChapman: `https://github.com/UKHomeOfficeForms/hof-model`

apps/common/models/file-upload.js
```diff
@@ -0,0 +1,33 @@
...
+          options: {
+            filename: this.get('name'),
+            contentType: this.get('mimetype')
+          }
+        }
+      };
+      reqConf.method = 'POST';
+      this.request(reqConf, (err, data) => {
+        if (err) {
+          reject(err);
```
JoeChapman: shouldn't you return to stop execution of `resolve`?

apps/new-dealer/controllers/supporting-documents-add-another.js
```diff
@@ -0,0 +1,27 @@
...
+    if (req.params.action === 'delete' && req.params.id) {
+      return this.removeItem(req, res);
+    }
+    return super.get(req, res, callback);
+  }
+
+  removeItem(req, res) {
+    const items = req.sessionModel.get('supporting-documents') || [];
+    req.sessionModel.set('supporting-documents', items.filter((file) => file.id !== req.params.id));
+    res.redirect(path.join(req.baseUrl, this.options.route));
```
JoeChapman: I don't think this is a suitable method for this line of code, because it has nothing to do with removing the item.

apps/new-dealer/controllers/supporting-documents.js
```diff
@@ -0,0 +1,41 @@
+'use strict';
+
+const BaseController = require('../../common/controllers/base');
+const UploadModel = require('../../common/models/file-upload');
+
+const uuid = require('node-uuid');
```
JoeChapman: use `uuid`, a fork of node-uuid, maintained more regularly and https://github.com/kelektiv/node-uuid/issues/116

config.js
```diff
@@ -4,7 +4,7 @@
 module.exports = {
   env: process.env.NODE_ENV,
   postcode: {
-    hostname: process.env.NODE_ENV === 'ci' ?
+    hostname: process.env.NODE_ENV === 'ci' || process.env.NODE_ENV === 'development' ?
```
JoeChapman: or `env === 'ci || env === 'development'`... which is possibly easier to read

config.js
```diff
@@ -4,7 +4,7 @@
 module.exports = {
```
JoeChapman: `const config = {...`

config.js
JoeChapman: `config.env === 'ci' || config.env === 'development' `

config.js
```diff
@@ -13,6 +13,11 @@ module.exports = {
       param: 'postcode'
     }
   },
+  upload: {
+    hostname: process.env.NODE_ENV === 'ci' || process.env.NODE_ENV === 'development' ?
```
JoeChapman: `config.env === 'ci' || config.env === 'development' `

config.js
```diff
@@ -13,6 +13,11 @@ module.exports = {
...
+      `http://${process.env.LISTEN_HOST || '0.0.0.0'}:${process.env.PORT || 8080}/api/file-upload` :
+      'TODO-PROVIDE-URL'
```
JoeChapman: this'll never work

package.json
```diff
@@ -33,18 +33,22 @@
...
   "dependencies": {
     "browserify": "^12.0.1",
-    "hof-bootstrap": "^6.0.0",
+    "busboy-body-parser": "^0.3.0",
+    "hof-bootstrap": "^8.0.1",
     "hof-frontend-toolkit": "^1.0.0",
     "hof-transpiler": "^0.1.0",
     "jquery": "^2.1.4",
     "node-sass": "^3.7.0",
+    "node-uuid": "^1.4.7",
```
JoeChapman: use `uuid`

test/common.js
```diff
@@ -1,7 +1,9 @@
 'use strict';
 
-global.chai = require('chai').use(require('sinon-chai'));
-global.should = chai.should();
```
JoeChapman: :(

apps/common/models/file-upload.js
lennym: I just copied this - https://github.com/UKHomeOffice/firearms/blob/master/apps/common/models/postcodes.js#L4

Happy to change both though.

apps/common/models/file-upload.js
lennym: It doesn't strictly matter because promises can only complete once, but I'll change it anyway for clarity.

apps/new-dealer/controllers/supporting-documents-add-another.js
lennym: How would you recommend structuring the code? It seems fine to me.

config.js
lennym: I know, but while there is no endpoint - @easternbloc is working on it - we need something, and a `TODO` that crashes and errors is better than something which fails silently.

test/common.js
lennym: Why so sad?

apps/new-dealer/controllers/supporting-documents-add-another.js
JoeChapman: I would prefer it if the loc that redirects were inside it's own method. It looks as if the redirection is a side effect of calling a method that I would expect to exclusively remove and item.

config.js
JoeChapman: no problem

JoeChapman: @easternbloc do you have a end point?

easternbloc: I will have once you lot merge a PR I'm about to raise

config.js
easternbloc: @JoeChapman what are you saying here?

config.js
lennym: Just spotted this. I'm curious to know why you would do `const config = {...}; module.exports = config;` over exporting the object directly.

What would you see as the benefit?

PR #103, opened 2017-01-20
Ensure translations are updated in dev mode when lang files change

lennym:
The previous setup meant that the lang files were compiled when files changed, but did not trigger a restart of the server, so it would have to be restarted manually in order to see text changes. By including the lang file compilation in the main app start-up we can ensure that the running app in dev mode always shows the most recent content.

This does cause a slight overhead for restarts due to regular code changes, but it is minimal.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,6 @@
     "watch:scss": "nodemon -e scss -x 'npm run sass'",
     "watch:js": "nodemon --watch assets/js -e js -x 'npm run browserify'",
     "test": "NODE_ENV=test npm run lint && istanbul cover _mocha",
-    "test:ci": "npm run test",
     "test:acceptance": "funkie so-acceptance --steps",
     "lint": "npm-run-all --parallel lint:app lint:acceptance",
     "lint:app": "eslint .",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -12,8 +12,8 @@
   "readme": "./README.md",
   "scripts": {
     "start": "node .",
-    "dev": "npm-run-all --parallel watch:app watch:scss watch:js watch:translations",
-    "watch:app": "NODE_ENV=development nodemon .",
+    "dev": "npm-run-all --parallel watch:app watch:scss watch:js",
+    "watch:app": "NODE_ENV=development nodemon -x 'npm run hof-transpile && npm start' --ignore 'apps/**/translations/en/'",
     "watch:scss": "nodemon -e scss -x 'npm run sass'",
     "watch:js": "nodemon --watch assets/js -e js -x 'npm run browserify'",
     "watch:translations": "nodemon --watch apps/**/translations/src -x 'npm run hof-transpile'",
```
joefitter: does this work? glob wildcards don't work in nodemon according to the docs - https://github.com/remy/nodemon#monitoring-multiple-directories

lennym: Dunno, that's not part of the diff. It can probably be removed though tbh since `watch:translations` is made redundant by this PR.

JoeChapman: it's not nodeman that parses the argument, hof-transpile does. 
Yeah, it works, but if it's redundnat now, meh...

lennym: Removed it.

PR #106, opened 2017-01-31
Refactor address lookup

lennym:
Address lookup was confusing me deeply. There were 8 different controllers available to perform seemingly 3 common form steps, the categories functionality that was specific to one instance in shooting clubs was baked into all of them, and I had no idea what was what.

I've tried to put together a refactor of addresses into a set of reusable controllers in a way that makes sense to me.

## Goals:

* Since address lookup functionality is dependent on field names, do not allow field names to be driven by field config since that means config changes can break functionality.
* Address data should end up with the same structure, irrespective of whether the values were created using manual entry or postcode lookup.
* Common aspects of config should be standardised and encapsulated into controllers.
* Move towards being able to extract address lookup as a common controller/set of controllers.
* Separate address functionality from looping steps. Looping and collecting results into a set should be independent from the data being collected.

This is very much work in progress, but I wanted to get a review of the overall ideas, before going too far into the detail. It's all been implemented and works well enough in the shooting clubs form, but I haven't started implementing for other forms yet.


lennym:
I've added a README to `./apps/common/controllers/address` which elaborates on the implementation a little.

apps/common/controllers/address/helper.js
```diff
@@ -0,0 +1,55 @@
...
+    'prefix',
+    'start',
+    'select',
+    'manual',
+    'next'
+  ];
+
+  required.forEach((key) => {
+    if (!config[key]) {
+      throw new Error(`options.${key} must be defined`);
```
JoeChapman: Should this be `config.${key}...`?

lennym: I'm going to take the fact that you're arguing the semantics of "config" vs "options" in error messaging as a good sign.

> I wanted to get a review of the overall ideas, before going too far into the detail

apps/shooting-clubs/index.js
```diff
@@ -137,134 +102,56 @@ module.exports = {
...
+      next: '/location-add-another-address'
     },
     '/location-add-another-address': {
       addressKey: 'locationAddresses',
       template: 'add-another-address-loop.html',
-      controller: require('../common/controllers/add-another-address-loop'),
-      fields: [
-        'location-add-another-address'
+      controller: require('./controllers/location-address-loop'),
+      next: '/storage-address',
```
sulthan-ahmed: I didn't quite understand storage-address

apps/shooting-clubs/index.js
```diff
@@ -137,134 +102,56 @@ module.exports = {
...
-      }
+      next: '/location-add-another-address'
     },
     '/location-add-another-address': {
       addressKey: 'locationAddresses',
       template: 'add-another-address-loop.html',
-      controller: require('../common/controllers/add-another-address-loop'),
-      fields: [
-        'location-add-another-address'
+      controller: require('./controllers/location-address-loop'),
```
sulthan-ahmed: what controller is it if you don't have a category? is it still location-address-loop?

sulthan-ahmed: my bad, I see you have add-another-address-loop.js

easternbloc:
@lennym crosses fingers

PR #112, opened 2017-02-03
Swap node-sass for npm-sass

lennym:
This makes it easier, more reliable and cleaner to import sass from child modules.

This isn't *just* me tying to add my own modules. If we standardise on npm-sass instead of node-sass then we can get rid of the *nasty* build step in `hof-frontend-toolkit` to resolve the install paths.

Also this:

screen shot 2017-02-03 at 10 50 07


sulthan-ahmed:
LOL at `This isn't just me trying to add my own modules.`

sulthan-ahmed:
As discussed with lenny as this is a small PR, I have reviewed and I'll merge

PR #119, opened 2017-02-09
Museums and Shooting Clubs confirm pages

lennym:


apps/common/controllers/confirm.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
+
+const controllers = require('hof-confirm-controller');
+
+module.exports = class ConfirmController extends controllers {
+
+  locals(req, res) {
+    const content = req.rawTranslate('pages.confirm');
```
lennym: I want there to be a better way of doing this, but it's taken from new-dealer for now, and I don't have the time to deal with it. ¯\_(ツ)_/¯

apps/museums/index.js
```diff
@@ -21,6 +21,7 @@ const contactAddressLookup = AddressLookup({
 module.exports = {
   name: 'museums',
   baseUrl: '/museums',
+  params: '/:action?',
```
lennym: This should be defaulted in bootstrap. It's in 100% of our apps.

apps/common/controllers/confirm.js
joefitter: yeah it sucks for 2 reasons, firstly because deepTranslate tries to lookup conditional values if a translation returns an object - this should be fixed in hof-middleware/deep-translate. Second reason as we discussed, we should look into markdown templates or similar. But also as you said, ¯\_(ツ)_/¯

apps/museums/index.js
joefitter: agree - we should still have the ability to add more params though

apps/shooting-clubs/controllers/confirm.js
```diff
@@ -0,0 +1,24 @@
...
+const Controller = require('../../common/controllers/confirm');
+
+module.exports = class ConfirmController extends Controller {
+
+  getFieldData(field, req) {
+    const defaults = super.getFieldData(field, req);
+
+    if (field.field === 'location-addresses') {
+      return defaults.value.map(address => {
+        address['address-category'] = [].concat(address['address-category']);
```
joefitter: do we need to mutate this? I don't know exactly what it's doing, so the answer is probably yes, but it feels kinda icky

lennym: If you only select a single checkbox then the value is a string. This makes sure it's always an array.

joefitter: but rather than actually mutating the value here, cant we just cast it below in the Object assign?

```js
return Object.assign({}, defaults, {
  label: address.address,
  value: _.castArray(address['address-category'])
    .map(category => req.translate(`fields.location-address-category.options.${category}.label`))
    .join('\n')
});
```

lennym: Sure, we could. I just don't see why it's better.

And with hindsight this normalisation of values should probably get done at the field format level. It's nasty that a field can either yield a string or an array depending on user input, and it should be fixed at source rather than where it has effects.

sulthan-ahmed:
not sure this is related to your PR lenny but when I run the app, I get 

`/Users/Sulthan/Development/projects/HO/firearms/node_modules/hof-transpiler/lib/compile.js:15
    throw new Error(error);
    ^

Error: SyntaxError: Unexpected end of JSON input
    at parseFileSync (/Users/Sulthan/Development/projects/HO/firearms/node_modules/hof-transpiler/lib/compile.js:15:11)
    at Walker.<anonymous> (/Users/Sulthan/Development/projects/HO/firearms/node_modules/hof-transpiler/lib/compile.js:91:39)
    at emitThree (events.js:116:13)
    at Walker.emit (events.js:194:7)
    at Object.emitSingleEvents [as emitNodeType] (/Users/Sulthan/Development/projects/HO/firearms/node_modules/walk/lib/node-type-emitter.js:55:13)
    at Walker._wLstatHandler (/Users/Sulthan/Development/projects/HO/firearms/node_modules/walk/lib/walk.js:87:19)
    at /Users/Sulthan/Development/projects/HO/firearms/node_modules/walk/lib/walk.js:105:12
    at FSReqWrap.oncomplete (fs.js:123:15)`

lennym:
@sulthan-ahmed What version of `hof-transpiler` do you have installed? The version updated to 0.2.1 in this PR.

lennym:
Tests are going to need more time than we have available. Do not expect them.

sulthan-ahmed:
:(

sulthan-ahmed:
In node modules it says  "version": "0.2.1"

apps/shooting-clubs/controllers/confirm.js
joefitter: just because if `super.getFieldData` returned a reference to an item, we would actually be changing this value in the loop. Totes agree that this should be field level formatting

package.json
```diff
@@ -35,13 +35,11 @@
     "busboy-body-parser": "^0.3.0",
     "express": "^4.14.1",
     "hof-bootstrap": "^9.0.1",
```
JoeChapman: can you update to 9.3.1 for CSP please

lennym: ^9.0.1 will install 9.3.1 already.

lennym:
https://github.com/UKHomeOfficeForms/hof-transpiler/pull/31 Not a fix for the transpiler error, but will make debugging easier.

sulthan-ahmed:
cool, I'll just do an npm i to update my hof-transpiler and see what that says

sulthan-ahmed:
```> hof-transpiler ./apps/**/translations/src -w --shared ./apps/common/translations/src --shared ./node_modules/hof-template-partials/translations

undefined:1
{


SyntaxError: Unexpected end of JSON input
    at Object.parse (native)
    at parseFileSync (/Users/Sulthan/Development/projects/HO/firearms/node_modules/hof-transpiler/lib/compile.js:12:15)
    at Walker.<anonymous> (/Users/Sulthan/Development/projects/HO/firearms/node_modules/hof-transpiler/lib/compile.js:87:39)
    at emitThree (events.js:116:13)
    at Walker.emit (events.js:194:7)
    at Object.emitSingleEvents [as emitNodeType] (/Users/Sulthan/Development/projects/HO/firearms/node_modules/walk/lib/node-type-emitter.js:55:13)
    at Walker._wLstatHandler (/Users/Sulthan/Development/projects/HO/firearms/node_modules/walk/lib/walk.js:87:19)
    at /Users/Sulthan/Development/projects/HO/firearms/node_modules/walk/lib/walk.js:105:12
    at FSReqWrap.oncomplete (fs.js:123:15)```

File: apps/shooting-clubs/controllers/confirm.js (hunk 3 above)
Comment (lennym) [inline apps/shooting-clubs/controllers/confirm.js]: I tried writing a formatter to do this, then noticed that even if I do, `hof-for-controller` immediately slaps be back down - https://github.com/UKHomeOfficeForms/hof-form-controller/blob/master/lib/formatting/index.js#L27-L29

Comment (lennym) [inline apps/shooting-clubs/controllers/confirm.js]: I've update the code as per your comment though.

File: package.json (hunk 4 above)
Comment (JoeChapman) [inline package.json]: sorry, I took a shot, and missed

File: apps/common/controllers/address/select.js [hunk 5]
```diff
@@ -6,6 +6,11 @@ module.exports = class AddressSelectController extends AddressController {
 
   configure(req, res, callback) {
     const addresses = req.sessionModel.get(`${req.form.options.prefix}-address-list`);
+    if (!addresses || !addresses.length) {
+      const err = new Error();
+      err.code = 'MISSING_PREREQ';
+      callback(err);
+    }
```
joefitter: 👍

PR #126, opened 2017-02-17
Add validation messages

lennym:
Includes a bootstrap bump to latest to get some code allowing modification of errors from latest hof-form-controller. This allows setting different error messages on a field that exists in two places with the same name - address manual entry vs selection.


lennym:
This looks to have caused a bunch of acceptance test failures when I run it locally, so hang fire on merging for now.

Feel free to review though.

easternbloc:
@lennym I updated the title to ensure nobody accidentally does a drive by

lennym:
I know what you're thinking...

> Wow, that's a lot of PR for some validation messages

Yes... because I had to major bump `hof-bootstrap` to get new error classes from `hof-form-controller` which broke a heap of stuff that had to be yak-shaved back together.

Should be ready to merge now.

apps/museums/translations/src/en/validation.json
```diff
@@ -0,0 +1,38 @@
...
+  },
+  "contact-postcode": {
+    "required": "Please enter a postcode",
+    "postcode": "Please enter a valid UK postcode"
+  },
+  "contact-address": {
+    "required": "Please select an address",
+    "required-manual": "Please enter an address"
+  }
+}
```
joefitter: I know you probably don't care - but generally we don't use please

apps/new-dealer/controllers/authority-holders-nationality.js
```diff
@@ -13,19 +13,19 @@ module.exports = class AuthorityHolderNationalityController extends BaseControll
       const thirdNationality = req.form.values[`${key}-third`].toLowerCase();
 
       if (firstNationality === secondNationality && firstNationality !== '') {
-        return new ErrorController(`${key}-second`, {
+        return new ValidationError(`${key}-second`, {
```
joefitter: can you not use `this.ValidationError`?

lennym: I was just renaming variables because `ErrorController` is a *bad* variable name, but you are right and I will do that thing you said anyway.

apps/museums/translations/src/en/validation.json
lennym: Well that's just... rude!

joefitter: ha I know, I like please, but apparently we don't tend to use it - cant remember the exact reason - maybe @Ben-Stevens can enlighten?

lennym: I've updated now anyway.

joefitter:
LGTM pending build

lennym:
I was silly enough to fix the acceptance tests, so build might take a while...

PR #133, opened 2017-03-06
Add step to collect number of shooting club members

lennym:


lennym:
This is blocked pending a decision on where the step should appear.

PR #134, opened 2017-03-09
Add behaviour for submission to iCasework

lennym:
Provides a behaviour that makes a POST to iCasework's create case API as part of `saveValues`. The behaviour allows for a custom `prepare` method to be passed as part of the behaviour configuration which can extend the default auth attributes with the appropriate case data for each of our applications.

```js
// in steps config
const CaseWorkSubmission = require('../common/behaviours/casework-submission');

'/confirm': {
  behaviours: [
    'complete',
    CaseWorkSubmission({
      prepare: (session) => { /* return data fields */ }
    })
  ],
  next: '/confirmation'
  ...
}
```

### DIFF
--- a/apps/common/behaviours/casework-submission.js
+++ b/apps/common/behaviours/casework-submission.js
@@ -1,0 +1,35 @@
+'use strict';
+
+const CaseworkModel = require('../models/i-casework');
+
+const Compose = func => superclass => class extends superclass {
+
+  prepare() {
+    if (typeof func === 'function') {
+      return Object.assign(super.prepare(), func(this.toJSON()));
+    }
+    return super.prepare();
+  }
+
+};
+
+module.exports = config => {
+
+  config = config || {};
+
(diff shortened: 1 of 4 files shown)


apps/common/behaviours/casework-submission.js
```diff
@@ -0,0 +1,35 @@
...
+  // compose custom per-app prepare methods onto the standard model
+  const Model = Compose(config.prepare)(CaseworkModel);
+
+  return superclass => class extends superclass {
+
+    saveValues(req, res, next) {
+      const model = new Model(req.sessionModel.toJSON());
+      model.save()
+        .then(() => {
+          super.saveValues(req, res, next);
```
lennym: I'm *guessing* we'll want to save the casework id returned and present it to the user somehow, but as with all things, requirementslol.

easternbloc: I'd assume on the confirmation and via email.

PR #136, opened 2017-03-09
Add happy path tests to all apps

lennym:


apps/museums/fields/index.js
```diff
@@ -15,7 +15,7 @@ module.exports = {
   },
   'contact-phone': {
     mixin: 'input-text',
-    validate: ['required', 'phonenumber']
+    validate: 'required'
```
lennym: Automated testing exposed that out of the box phone number validation fails with spaces in the number. Options are:

* change the validator
* format the value to remove spaces
* remove the validator

Since other apps in firearms don't validate phone numbers, I went for the quickest option.

JoeChapman: How would you want to present the phone number back to the user (assuming you will need to do so on the confirm page)? If you need/would like to present it as it was input, formatting would need to occur twice. I vote for changing the formatter

apps/supporting-documents/acceptance/features/_happy-path.js
```diff
@@ -0,0 +1,33 @@
...
+
+// create a function that returns the next item of an array every time it is called
+const iterate = (arr) => {
+  let count = 0;
+  return () => {
+    const value = arr[count];
+    count++;
+    return value || arr[arr.length - 1];
+  };
+};
```
JoeChapman: This function looks familiar

JoeChapman:
LGTM

JoeChapman:
Anyone for anymore? @joefitter

apps/supporting-documents/acceptance/features/_happy-path.js
lennym: You've seen it before, that's why.

I've tried thinking of a good place for it to move it, without joy. Mostly because I'd like to make it automatic behaviour in the auto-filler, but can't come up with a nice way of handling it that is safely different from using arrays to specify multiple checkbox selections.

While I ponder on a *good* abstraction I figure no abstraction is better than a bad one. It's not exactly a huge amount of code.

apps/museums/fields/index.js
lennym: Changing the formatter wasn't an option.

`s/formatter/validator`?

I'm inclined to agree, but it's buried right down in `hof-form-controller` so this was a pragmatic fix in the short/medium term.

easternbloc:
happy to merge when drone is done

PR #137, opened 2017-03-13
Feature/pdf file

easternbloc:
This is available on musesums and shooting clubs atm

We will enable it on s5 when it's been migrated to the new confirm controller

### DIFF
--- a/apps/common/behaviours/generate-pdf.js
+++ b/apps/common/behaviours/generate-pdf.js
@@ -26,13 +26,13 @@ module.exports = superclass => class extends superclass {
     });
   }
 
-  generatePDF(req, res, next) {
+  generatePDF(req, res, callback) {
     const locals = Object.assign({}, this.locals(req, res), {
       title: 'Firearms Application'
     });
     res.render('pdf.html', locals, (err, html) => {
       if (err) {
-        return next(err);
+        return callback(err);
       }
 
       // phantom is weird about paths it doesn't read things like a webserver
@@ -40,12 +40,7 @@ module.exports = superclass => class extends superclass {
       pdf.create(html, {
         base: 'file://' + path.resolve(__dirname, '../../../') + '/',
         border: '1cm'
-      }).toBuffer((e, buffer) => {
-        if (e) {
-          return next(e);
-        }
-        next(null, buffer);
-      });
+      }).toBuffer(callback);
     });
   }
 };
--- a/apps/common/behaviours/generate-pdf.js
+++ b/apps/common/behaviours/generate-pdf.js
@@ -2,12 +2,14 @@
 
 const pdf = require('html-pdf');
 const UploadModel = require('../../common/models/file-upload');
-const concat = require('concat-stream');
 const path = require('path');
 
 module.exports = superclass => class extends superclass {
   process(req, res, next) {
-    this.generatePDF(req, res, (pdfBuffer) => {
+    this.generatePDF(req, res, (err, pdfBuffer) => {
+      if (err) {
+        return next(err);
+      }
       const file = {
         name: 'application_form.pdf',
         data: pdfBuffer,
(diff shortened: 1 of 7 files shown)


apps/common/behaviours/generate-pdf.js
```diff
@@ -0,0 +1,53 @@
+'use strict';
+
+const pdf = require('html-pdf');
+const UploadModel = require('../../common/models/file-upload');
+const concat = require('concat-stream');
+const path = require('path');
+
+module.exports = superclass => class extends superclass {
+  saveValues(req, res, callback) {
+    super.saveValues(req, res, (err, values) => {
```
lennym: I'd prefer if all of our behaviours did their thing and *then* called super. It means we have a predictable order of operations when composing behaviours.

lennym: Specifically in this case because one of the other behaviours with a `saveValues` will be to submit the application to iCasework, and that definitely needs to happen *after* this has completed.

lennym: In fact, the more I think about this, the more I think it should be `process` and not `saveValues` for exactly that reason. 

`process` is generally where we write to `req.form.values`

package.json
```diff
@@ -42,6 +43,7 @@
     "hof-middleware-markdown": "^1.0.0",
     "hof-model": "^3.0.0",
     "hof-transpiler": "^0.2.1",
+    "html-pdf": "^2.1.0",
```
lennym: This creates a dependency on `phantomjs-prebuilt` which might cause problems with our policy of not allowing install scripts to run in containers, because it has one.

assets/scss/app.scss
```diff
@@ -209,3 +209,29 @@ div.add-remove-section {
...
+  footer,
+  #declaration-section,
+  #skiplink-container {
+    display: none;
+  }
+}
+
+html {
+  background: white;
+}
```
JoeChapman: This should be a separate stylesheet, imported into `app.scss`

JoeChapman:
do you intend for this to ultimately run as a separate service in it's own container?

apps/common/behaviours/generate-pdf.js
```diff
@@ -0,0 +1,50 @@
...
+      title: 'Firearms Application'
+    }), (err, html) => {
+      if (err) {
+        next(err);
+      }
+      let concatStream = concat(next);
+      // phantom is weird about paths it doesn't read things like a webserver
+      html = html.replace(/href="\//g, 'href="./');
+      pdf.create(html, {
+        base: 'file://' + path.resolve(__dirname, '../../../') + '/',
```
JoeChapman: Presuming this points to the folder containing, images, css and js. It should be local to the behaviour, not the <project>/assets and scss compiled dynamically if necessary.

assets/scss/app.scss
lennym: I'd take the opposite approach. The pdf stylesheet should be a separate stylesheet altogether, which imports app.scss.

apps/common/behaviours/generate-pdf.js
easternbloc: yes if it was a global behaviour for use outside of firearms but it's not so meh

assets/scss/app.scss
easternbloc: 🍆

apps/common/behaviours/generate-pdf.js
easternbloc: 🍆

JoeChapman: 🍅 🍆  🍅

lennym: 🍒🍆

apps/common/behaviours/generate-pdf.js
```diff
@@ -0,0 +1,48 @@
...
+        }, next);
+    });
+  }
+
+  generatePDF(req, res, next) {
+    res.render('pdf.html', Object.assign({}, this.locals(req, res), {
+      title: 'Firearms Application'
+    }), (err, html) => {
+      if (err) {
+        next(err);
```
lennym: `return next(err);`

apps/common/behaviours/generate-pdf.js
```diff
@@ -0,0 +1,48 @@
...
+      }
+      let concatStream = concat(next);
+      // phantom is weird about paths it doesn't read things like a webserver
+      html = html.replace(/href="\//g, 'href="./');
+      pdf.create(html, {
+        base: 'file://' + path.resolve(__dirname, '../../../') + '/',
+        border: '1cm'
+      }).toStream((e, stream) => {
+        if (e) {
+          next(e);
```
lennym: `return`

apps/common/behaviours/generate-pdf.js
```diff
@@ -0,0 +1,51 @@
...
+      html = html.replace(/href="\//g, 'href="./');
+      pdf.create(html, {
+        base: 'file://' + path.resolve(__dirname, '../../../') + '/',
+        border: '1cm'
+      }).toBuffer((e, buffer) => {
+        if (e) {
+          return next(e);
+        }
+        next(null, buffer);
+      });
```
lennym: At this point you could just not bother with the extra function and just `.toBuffer(next)`. No biggie though.

lennym: As an aside, I also try to avoid ever using `next` for callbacks that are *not* express middleware callbacks, and always use `callback` for those. That way I have some consistency of expectation.

`next` => express middleware, takes no params but an error
`callback` => regular, old-fashioned callback, can receive a second parameter

PR #138, opened 2017-03-16
Unset manual/lookup address fields when completing the other

lennym:
If a user goes through an address journey more than once, following manual entry *and* a successful lookup on different occasions then they end up with both fields populated on the confirm page, and two addresses will appear.

Instead unset the other address field whenever either a manual address or a lookup is completed.


lennym:
Realised there's a better way to do this directly with config. Closing.

PR #140, opened 2017-03-23
Submission models

lennym:
This needs testing end-to-end with file uploads and icasework, but this is a decent enough first candidate.

Each app has `apps/**/models/submission.js` which returns mapped fields for that app. Default stuff and auth is handled in `apps/common/models/i-casework.js`.

To test this, you'll need icasework auth params as environment variables. Hit me up for these.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -44,36 +44,36 @@ pipeline:
     when:
       event: pull_request
 
-  build_acceptance:
-    image: docker:1.11
-    environment:
-      - DOCKER_HOST=tcp://127.0.0.1:2375
-    commands:
-      - docker build -f Dockerfile-acceptance -t acceptance .
-    when:
-      branch: master
-      event: push
+  # build_acceptance:
+  #   image: docker:1.11
+  #   environment:
+  #     - DOCKER_HOST=tcp://127.0.0.1:2375
+  #   commands:
+  #     - docker build -f Dockerfile-acceptance -t acceptance .
+  #   when:
+  #     branch: master
+  #     event: push
 
-  test_acceptance:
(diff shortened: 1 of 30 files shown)


.drone.yml
```diff
@@ -44,36 +44,36 @@ pipeline:
...
+  # test_acceptance:
+  #   image: docker:1.11
+  #   environment:
+  #     - DOCKER_HOST=tcp://127.0.0.1:2375
+  #   commands:
+  #     - docker run -e NODE_ENV=ci -e DISABLE_CSP=true -d --net host firearms
+  #     - docker run --rm --net host acceptance
+  #   when:
+  #     branch: master
+  #     event: push
```
sulthan-ahmed: any reason why these are commented out rather than not deleted?

apps/common/views/partials/confirmation-alert.html
```diff
@@ -0,0 +1,8 @@
+<div class="alert-complete" role="alert">
+  <div class="alert-message">
+    <h1>{{#t}}pages.confirmation.alert{{/t}}</h1>
+    {{#values.caseid}}
+    <p>{{#t}}pages.confirmation.reference{{/t}}</p>
+    {{/values.caseid}}
+  </div>
+</div>
```
sulthan-ahmed: missing EOF line

.drone.yml
easternbloc: they will return in the next few days so i'm just commenting them out for now

lennym: Bet they don't

sulthan-ahmed: that's fine

PR #160, opened 2017-04-13
Use hof-build module for asset compilation

lennym:
This makes the package.json *much* cleaner, and provides a massive performance boost to runnign in development mode.


Dockerfile
```diff
@@ -11,7 +11,7 @@ RUN yum clean all && \
 COPY package.json /app/package.json
 RUN npm --loglevel warn install --production
 COPY . /app
-RUN npm --loglevel warn run postinstall
+RUN npm --loglevel warn run postinstall --production
```
lennym: This gives us code minification

PR #180, opened 2017-05-03
Editing loops

lennym:
There were a number of issues with editing things which are collected in a loop - e.g. addresses, supporting documents

* When you return to the "add another" step to edit, selecting "yes" just bumps you back to `/confirm`
* Deleting an entry breaks out of edit journey.
* Selecting manual address input breaks out of edit journey

These have now been fixed for addresses in S5 and shooting clubs, and supporting document uploads.

Also, some lingering code from when individual addresses could be edited (as opposed to just deleted and re-added) has been removed to clean things up.


sulthan-ahmed:
This looks good, I just wanted to run the app to check all of it works as you suggested.  However, I get the following when I do npm run dev 
`throw new Error('At least one of `from` or `replyTo` options must be defined');`

lennym:
@sulthan-ahmed https://github.com/UKHomeOffice/firearms/blob/v1.0.0/config.js#L32

sulthan-ahmed:
Thanks Lenny, it works now

So I tested the functionality and here is a summary with a question 
* s5
- clicking yes now works as per described
- deleting now works properly
- when you enter a manual entry then it works as normal.  In the dev version I couldn't get into the manual entry section, it would just take me to the confirm page but your PR works fine

* shooting clubs
- deleting works as per expectation
- manual entry works fine
- clicking yes, I think works normally but can I just confirm if this is working as expected?
If I click change on the shooting ranges, click yes and add a new address and the weapon category, then I am taken to the question, 'At which of these addresses, if any, will the firearms be stored?'.  I was initially expecting to go to the confirm page but I can understand why I would go to this question as there is a new address for storage

lennym:
@sulthan-ahmed  Some questions about your questions:

### S5

* "In the dev version" do you mean the deploye dev server, or localhost running this branch.
* What is your actual question about S5? You've just told me it all works fine.

### Shooting Clubs

* Clicking "yes" for what question?
* What happens _after_ you submit the "At which of these addresses.. [etc]" step? If it takes you to confirm then it's probably ok, if you have to re-fill all the fields then it's probably not.

sulthan-ahmed:
##s5
* @lennym it was just a summary to say that s5 is working fine, sorry I wasn't so clear. 
* It was on the deploy dev. I was just saying that I couldn't recreate the issue with entering a manual address.  I don't think it matters because it's working all fine

##Shooting club
Shooting clubs all works fine except one part which I'm not sure if it's what you want.  I'll try to explain
- You fill out the form and get to the confirm page
screen shot 2017-05-04 at 13 43 34

- You click change on the Shooting range address
- You click Yes on the question, `Will the club use another shooting range?`
screen shot 2017-05-04 at 13 48 14
- You add an address 
- You add a firearms category
- You then Say No on the question, `Will the club use another shooting range?`
- You are then presented with, `At which of these addresses, if any, will the firearms be stored?`
screen shot 2017-05-04 at 13 50 24
- Is this what you want? I was expecting to go to the confirm page at this point

Hope it now makes more sense what I'm saying

lennym:
That was a bug where clicking the manual address link lost the `/edit` parameter from the URL. Latest commit resolved that.

PR #182, opened 2017-05-05
Remove typeahead functionality from import field

lennym:
The user needs to be able to fill in more than one country of origin, or be able to provide other free-form notes.


JoeChapman:
Could we not present a standard select box with `multiple` attribute?

lennym:
We could, but that's not what's been asked for (also I don't think they're great from a UX perspective).

Also I think the intention is that the user can add explanatory notes if needed, and not *just* a list of countries.

PR #185, opened 2017-05-05
Add a link and complementing copy for survey links to confirmation pa…

JoeChapman:
…ge of each journey

### DIFF
--- a/apps/common/translations/src/en/pages.json
+++ b/apps/common/translations/src/en/pages.json
@@ -27,6 +27,9 @@
   "confirmation": {
     "alert": "Application complete",
     "reference": "Your reference number is {{values.caseid}}",
-    "email": "We have sent an email to {{values.contact-email}} to confirm receipt of your application."
+    "email": "We have sent an email to {{values.contact-email}} to confirm receipt of your application.",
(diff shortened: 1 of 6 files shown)


lennym:
I don't think translations is the right place for the survey urls. I'd move them to config.

lennym:
This would be better implemented as a behaviour than a controller. By creating a controller that extends from the core controller module then locals that are set in the [application-wide base controller](https://github.com/UKHomeOffice/firearms/blob/master/apps/common/controllers/base.js) are not included since that controler is no longer in the inheritance/prototype chain.

By implementing as a behaviour then the parent controller and the inheritance chain is not affected, as it is composed onto whatever underlying controllers are in place.

JoeChapman:
Ok, never written a behaviour before, I'll give it a shot though. The words you used appear to conspire into a type of logic so I'll update on Monday

lennym:
There's some docs for behaviours at https://ukhomeofficeforms.github.io/hof-guide/documentation/#behaviours which should (hopefully!) be helpful. Would appreciate feedback/improvements on the docs if there's anything that you think would help.

PR #198, opened 2017-05-18
Integrate html-pdf service

JoeChapman:
Refactor pdf upload behaviour to make request to html -> pdf service (ukhomeoffice/html-pdf-converter)

**This should not be merged without the relevant kube-firearms config**


apps/common/models/pdf.js
```diff
@@ -0,0 +1,24 @@
...
+    try {
+      JSON.parse(response.body || '{}');
+      let err = new Error();
+      err.body = response.body;
+      err.status = response.statusCode;
+      return callback(err, null, response.statusCode);
+    } catch (err) {
+      this.parseResponse(response.statusCode, response.body, callback);
+    }
+  }
```
lennym: I don't really get this function. It seems to throw an error when there isn't an error, and not throw an error when there is.

What is the intention?

---

Edit: I *think* I get it now. If it's a PDF (as we expect) then `JSON.parse` will throw and we end up in the catch block. Whereas errors will respond with JSON, so we know that if we have JSON then an error has occurred. I'm not sure I like it though. It's super confusing to have the success/error states apparently backwards like that.

config.js
```diff
@@ -30,11 +31,11 @@ module.exports = {
...
   },
   email: {
-    from: process.env.FROM_ADDRESS || '',
+    from: process.env.FROM_ADDRESS || 'me@stuff.com',
     transport: 'ses',
     transportOptions: {
-      accessKeyId: process.env.HOF_SES_USER || process.env.AWS_USER || '',
-      secretAccessKey: process.env.HOF_SES_PASSWORD || process.env.AWS_PASSWORD || ''
+      accessKeyId: process.env.HOF_SES_USER || process.env.AWS_USER || 'hgjhgkjh',
+      secretAccessKey: process.env.HOF_SES_PASSWORD || process.env.AWS_PASSWORD || 'hgjhgkjhgkjh'
```
lennym: Cool config bro.

JoeChapman: ahem

easternbloc:
@JoeChapman what about the kube files? Are you gonna drop in a pr for that also?

lennym:
I looks like we're using some `Buffer` methods that are only in node >= 6 so travis is failing on node 4.

This is fine since we're shipping on node 6, it just needs the travis.yml to be updated to ditch node 4.

apps/common/models/pdf.js
```diff
@@ -0,0 +1,23 @@
...
+
+  url() {
+    return config.pdf.url;
+  }
+
+  handleResponse(response, callback) {
+    if (isPdf(Buffer.from(response.body))) {
+      return this.parseResponse(response.statusCode, response.body, callback);
+    }
+    let err = new Error();
```
lennym: Should this be `const`?

lennym: The error should probably also have a message.

mocks/pdf.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+const router = require('express').Router();
+
+router.post('/', (req, res, next) => {
+  if (req.body.template) {
+    res.setHeader('Content-Type', 'octet-stream');
+    res.status(201).send(Buffer(100));
```
lennym: Does this buffer need to be modified to not fall foul of the `is-pdf` check?

config.js
```diff
@@ -51,5 +52,8 @@ module.exports = {
       'shooting-clubs': 'https://eforms.homeoffice.gov.uk/outreach/Feedback.ofml?FormName=shootclub/',
       museums: 'https://eforms.homeoffice.gov.uk/outreach/Feedback.ofml?FormName=Museums/'
     }
+  },
+  pdf: {
+    url: `http://${pdfUrl()}/convert` || `http://${localhost()}/api/pdf-convert`
```
lennym: This will never resolve to the mock. The first string is *always* truthy.

mocks/pdf.js
lennym: It might be easier just to have a dummy pdf file that gets sent back with `res.sendFile`

config.js
JoeChapman: ah yes

easternbloc:
@lennym comments?

config.js
```diff
@@ -51,5 +51,10 @@ module.exports = {
       'shooting-clubs': 'https://eforms.homeoffice.gov.uk/outreach/Feedback.ofml?FormName=shootclub/',
       museums: 'https://eforms.homeoffice.gov.uk/outreach/Feedback.ofml?FormName=Museums/'
     }
+  },
+  pdf: {
+    url: env !== 'production' ?
+    `http://${process.env.PDF_CONVERTER_HOST}:${process.env.PDF_CONVERTER_PORT}/convert` :
+    process.env.PDF_CONVERTER_URL
```
lennym: I'm not sure I grasp the reasoning behind using a different set of environment variables for production and "not production".

easternbloc: he's doing a bit of a dance for docker-compose there.

In prod we have another nginx with tls so no need for the port.

So you're right he should refactor the `docker-compse` template to do it there instead.

apps/common/behaviours/pdf-upload.js
```diff
@@ -0,0 +1,52 @@
...
+      })
+      .then(() => {
+        super.process(req, res, next);
+      }, next)
+      .catch((err) => {
+        next(err);
+      });
+  }
+
+  _render(req, res) {
```
lennym: Not super-thrilled about using `_render` as a method name because of `render` already being a distinct part of the request pipeline, and the leading underscore being an established hof pattern for private versions of extensible methods. It leads to the incorrect impression that this is related to the existing `render`phase of the request pipeline.

It *probably* doesn't matter *too much* for firearms, but if it's easy to update I'd prefer `renderPDF` (and possibly equivalent `createPDF` and `uploadPDF` methods).

easternbloc: `renderPDF` is nicer

config.js
lennym: It possibly looks like this is for docker-compose, but could we not just configure docker compose to have `PDF_CONVERTER_URL=http://html-pdf-converter:8001/convert` (since that's what it resolves to anyway) and dispense with the need for conditionals here?

JoeChapman:
Soooooo, who's going to merge this?

PR #202, opened 2017-05-24
Make the app work out the box with docker-compose

easternbloc:
Use our version of redis in docker-compose

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -12,6 +12,9 @@ services:
       - EMAIL_PORT=25
       - REDIS_HOST=redis
       - REDIS_PORT=6379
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -12,6 +12,9 @@ services:
       - EMAIL_PORT=25
       - REDIS_HOST=redis
       - REDIS_PORT=6379
+      - FROM_ADDRESS=caseworker@gov.uk.test
+      - AWS_USER=blah
+      - AWS_PASSWORD=passwordblah
```
JoeChapman: Exactly, what are the env vars actually doing?

PR #208, opened 2017-05-25
Use theme assets

lennym:
Instead of laoding hof-frontend-toolkit in as a separate dependency


JoeChapman:
moar rebase

assets/scss/app.scss
```diff
@@ -1,8 +1,7 @@
 $path: "../img/";
 
-@import "hof-frontend-toolkit";
+@import "hof-theme-govuk";
 
-@import "./alert.scss";
 @import "./icons.scss";
 @import "./validation.scss";
 @import "./print.scss";
```
JoeChapman: Are these stylesheets still required, are they not common?

lennym: They might be, I don't actually know. 

I just know that this particular PR doesn't make them any more common than they previously were.

`alerts.scss` was definitely rolled into `hof-frontend-toolkit` a few weeks ago (by me) so I removed that one, but I don't know the status of the others. I'll take a look.

lennym:
Re-re-based.

assets/scss/app.scss
lennym: I had a look and there's no duplication of code between here and `hof-frontend-toolkit`, so yes, these stylesheets are still required.

PR #216, opened 2017-05-31
Sign the amended yaml

JoeChapman:
....because a previous change to the drone file was not signed


lennym:
Is this supposed to fix the drone build?

easternbloc:
@lennym it would appear it now is yes.

JoeChapman:
@lennym @easternbloc yes, sorry I didn't make that clear. This is supposed to fix the failing build

PR #222, opened 2017-06-07
Remove old PDF converter dependencies

JoeChapman:


package.json
```diff
@@ -34,15 +34,13 @@
     "hof-model": "^3.1.0",
     "hof-theme-govuk": "^2.0.3",
     "hof-util-countries": "^1.0.0",
-    "html-pdf": "^2.1.0",
     "i18n-lookup": "^1.0.0",
     "is-pdf": "^1.0.0",
-    "jquery": "^2.1.4",
+    "jquery": "^3.2.1",
```
easternbloc: does the typeahead still work when you bump this?

JoeChapman: Checking

JoeChapman: Yes, it works.

PR #224, opened 2017-06-12
Remove form-label-bold class from the custom css

sulthan-ahmed:
* Previously by default all fields had a form-label-bold class by default.  The custom css was used to remove the bold.  This is no longer required because the latest hof-template-mixins fixes this issue
* The form-label-bold can now also be used for it's intended purpose

### DIFF
--- a/assets/scss/app.scss
+++ b/assets/scss/app.scss
@@ -38,10 +38,6 @@ h1 {
   @include bold-24;
 }
 
-.form-label-bold {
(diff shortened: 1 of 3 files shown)


lennym:
Could you PR a change to `hof` first that makes the version bump explicit please? And then require the new hof version here.

sulthan-ahmed:
makes sense @lennym

PR #226, opened 2017-06-14
Integrate the supporting docs app with icasework APIs

lennym:
* Adds an email step to collect the email address associated with the original application.
* Performs a lookup of the case data from icasework to validate the email address
* Adds the documents to the case in icasework
* Sends an email to the original email address if successful

Note: this will need a config change on kube-firearms to update the icasework url to remove the path.

### DIFF
--- a/apps/common/behaviours/casework-submission.js
+++ b/apps/common/behaviours/casework-submission.js
@@ -17,8 +17,11 @@ module.exports = config => {
 
   config = config || {};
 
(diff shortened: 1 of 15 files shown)


JoeChapman:
This looks legit. Are there any tests? Or if not, steps to reproduce?

lennym:
Merge this first: https://github.com/UKHomeOffice/kube-firearms/pull/57

PR #232, opened 2017-06-19
Add 500.html template to be rendered on 500 error

JoeChapman:
- make use of the latest error middleware which will
lookup your local template based on the status code.

- use the latest middleware to render 500 template
so server errors can use a different button text. ("Try again")

Depends on https://github.com/UKHomeOfficeForms/hof-middleware/pull/26


lennym:
I think this is redundant if https://github.com/UKHomeOfficeForms/hof-template-partials/pull/27 is merged, and I think it makes more sense to be done there as a default.

PR #236, opened 2017-06-21
Update nodejs to the latest LTS version

easternbloc:


lennym:
This doesn't actually affect our kubernetes deployments does it?

PR #240, opened 2017-06-29
Log error header on icasework failure

lennym:
icasework don't send anything useful in the response body for submission failures, so instead log the custom header they use for error messaging.


easternbloc:
this was on my list of todays things but yay for it already being done

PR #242, opened 2017-06-29
Set `AuthorityType` parameter on S5 iCasework submission

lennym:
The authority type parameter should be set according to the business activity of the applicant. Depending on the usage of the weapons they authirty will be classified as a dealer, carrier, dealer/carrier, or maritime guards.

### DIFF
--- a/apps/new-dealer/models/submission.js
+++ b/apps/new-dealer/models/submission.js
@@ -1,13 +1,13 @@
 'use strict';
 /* eslint complexity: 0 max-statements: 0 */
-const contains = (arr, val) => arr.indexOf(val) > -1 ? 'Yes' : 'No';
+const contains = (arr, val) => arr.includes(val) ? 'Yes' : 'No';
 
 const authorityType = usage => {
-  if (usage.indexOf('arm-guards') > -1) {
+  if (usage.includes('arm-guards')) {
     return 'Maritime Guards';
   }
 
-  if (usage.indexOf('transport') > -1 || usage.indexOf('transfer') > -1) {
+  if (usage.includes('transport') || usage.includes('transfer')) {
     // check if any other values are selected
     if (usage.filter(use => use !== 'transport' && use !== 'transfer').length) {
       return 'Carriers and Dealers';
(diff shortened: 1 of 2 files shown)


apps/new-dealer/models/submission.js
```diff
@@ -2,10 +2,27 @@
 /* eslint complexity: 0 max-statements: 0 */
 const contains = (arr, val) => arr.indexOf(val) > -1 ? 'Yes' : 'No';
 
+const authorityType = usage => {
+  if (usage.indexOf('arm-guards') > -1) {
+    return 'Maritime Guards';
+  }
+
+  if (usage.indexOf('transport') > -1 || usage.indexOf('transfer') > -1) {
```
JoeChapman: Just a suggestion, but there is a lot of repetition here. IMO the `indexOf` checks would be nicer abstracted to a helper function.

lennym: I don't think I agree, abstracting this kind of logic into a "helper" function just means you end up having to go and look somewhere else to work out what the actual logic is, and if anything makes the code harder to follow.

JoeChapman: I think you're right, going to look somewhere else to decipher the logic would not be a better idea. But perhaps `lodash#_.includes` might improve the readability. I tend to find several instances of `indexOf ... > -1` difficult to grok at a glance. That said, I am not super precious about this, just want to keep it simple.

lennym: TBH, since we're deploying on `node@6.9.x` I'd just go for `usage.includes('...')`.

I should get into the habit of doing that.

JoeChapman: sure, that's probably sensible

easternbloc:
Are you going to make those changes @lennym ?

lennym:
Not right now. I'm in the middle of the BRP migration. Feel free to either merge, change it yourself, or wait.

easternbloc:
Seeing as it's the highest importance thing for us right I'll update it.

PR #245, opened 2017-07-11
Add triple braces for values being passed in translations e.g {{{values.club-secretary-name}}}

sulthan-ahmed:
All variables are HTML escaped by default. This will return unescaped HTML to resolve the issue of apostrophes not displaying properly or any other such ASCII characters

Strangely the headers don't need the triple braces

https://jira.digital.homeoffice.gov.uk/browse/FLHO-704

### DIFF
--- a/apps/shooting-clubs/translations/src/en/pages.json
+++ b/apps/shooting-clubs/translations/src/en/pages.json
@@ -10,7 +10,7 @@
   },
   "second-contact-name": {
     "header": "Who should we contact in {{values.club-secretary-name}}'s absence?",
-    "intro": "We need to collect the contact details of a second person, in case {{values.club-secretary-name}} is unavailable"
+    "intro": "We need to collect the contact details of a second person, in case {{{values.club-secretary-name}}} is unavailable"
   },
   "second-contact-email": {
     "header": "What are {{values.second-contact-name}}'s contact details?"


apps/shooting-clubs/translations/src/en/pages.json
```diff
@@ -10,7 +10,7 @@
   },
   "second-contact-name": {
     "header": "Who should we contact in {{values.club-secretary-name}}'s absence?",
```
sulthan-ahmed: For some reason it doesn't need the triple braces and it displays properly

JoeChapman: What name were you testing it with?

sulthan-ahmed: Jon O'Donnell

JoeChapman: is that the same name that was causing an issue in the other context?

sulthan-ahmed: Yes pretty much, vijay wrote Jon O'Donneil.  I tested that too and that has the the same behaviour.

JoeChapman: I would need to see precisely what Vijay and you are entering. It could be that Vijay used a back tick or pasted the name into the field from a formatted text. Anyway, I'll merge this and Vijay can watch out for other problems

sulthan-ahmed: cool, I could have just added triple braces for all situations but since it wasn't coming back with any issues, I left it

PR #248, opened 2017-07-20
Install bzip2 on docker build

lennym:
bzip2 is required to successfully install phantomjs because if a binary is not found locally it downloads and unzips one.


lennym:
Should we actually put this in the nodejs base image, since installing phantomjs is likely to be a common thing we want to do?

easternbloc:
oopsie!

PR #255, opened 2017-11-06
Coerce `usage` to an array before filter

JoeChapman:
Because `usage` can be either an array or a string this amend concats `usage` to an array, so that `filter()` can be invoked

This was throwing unhandled errors and causing the app not to respond


JoeChapman:
@lennym unit test added

lennym:
I don't have access anymore apparently.

JoeChapman:
And I don't have admin rights @easternbloc

PR #256, opened 2017-11-20
Throw timeout errors when calling icasework apis

lennym:
Failures of IP whitelisting with icasework manifest with a long pause (request takes ~120s) followed by an nginx "bad gateway" error due to it hitting its default timeout.

By setting an explicit timeout on our calls to icasework we can send a timeout error page before nginx fails, and so prevent the nginx error page from displaying.

Additionally we prevent catching the error when calling `getcasedetails` and so allow the app to fail early - rather than allowing the user to continue and then displaying an incorrect credentials error later. Since the timeout error is not related to sending a good or bad reference this does not represent a security issue.

### DIFF
--- a/apps/common/models/i-casework.js
+++ b/apps/common/models/i-casework.js
@@ -7,6 +7,11 @@ const config = require('../../../config');
 
 module.exports = class CaseworkModel extends Model {
 
+  constructor(attributes, options) {
+    super(attributes, options);
+    this.options.timeout = this.options.timeout || 20000;
+  }
+
   url() {
     return config.icasework.url + config.icasework.createpath;
   }
--- a/apps/common/models/i-casework.js
+++ b/apps/common/models/i-casework.js
@@ -9,7 +9,7 @@ module.exports = class CaseworkModel extends Model {
 
   constructor(attributes, options) {
     super(attributes, options);
-    this.options.timeout = this.options.timeout || 20000;
+    this.options.timeout = this.options.timeout || config.icasework.timeout;
   }
 
   url() {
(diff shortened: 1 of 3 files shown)


apps/common/models/i-casework.js
```diff
@@ -7,6 +7,11 @@ const config = require('../../../config');
 
 module.exports = class CaseworkModel extends Model {
 
+  constructor(attributes, options) {
+    super(attributes, options);
+    this.options.timeout = this.options.timeout || 20000;
```
JoeChapman: any particular reason why you chose `20000` as the default? 
What is the default nginx timeout?

lennym: Not really - it's a bit arbitrary.

From what I can work out the default in nginx is 60s, but I don't know that for certain.

On reflection, this should be made a config parameter. I'll do that in the morning.

lennym: Citation: http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_read_timeout

lennym:
I've moved the setting of the timeout to config. This should be good to merge now.

config.js
```diff
@@ -41,12 +41,13 @@ module.exports = {
     password: process.env.REDIS_PASSWORD
   },
   icasework: {
-    url: process.env.ICASEWORK_URL || 'https://uat.icasework.com',
+    url: 'http://localhost:5678',
```
JoeChapman: um?!

lennym: Well... that was an error. Will fix.

lennym:
Fixed.

PR #261, opened 2018-01-25
[FLHO-732] - error message for invalid nationality

sulthan-ahmed:
this is when a user does not select a country on a list for the second nationality and third nationality for the first authority holder's nationality

### DIFF
--- a/apps/new-dealer/translations/src/en/validation.json
+++ b/apps/new-dealer/translations/src/en/validation.json
@@ -151,9 +151,11 @@
     "equal": "Enter a nationality from the list that appears"
   },
   "first-authority-holders-nationality-second": {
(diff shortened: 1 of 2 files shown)


apps/new-dealer/translations/src/en/validation.json
```diff
@@ -151,9 +151,11 @@
     "equal": "Enter a nationality from the list that appears"
   },
   "first-authority-holders-nationality-second": {
+    "equal": "Enter a nationality from the list that appears",
     "same-first-nationality": "Second nationality cannot be the same as the first nationality"
```
JoeChapman: Is there a test to assert there is an error (don't assert on the error message, just that it's there) when the nationality is invalid?

PR #269, opened 2018-05-24
Ensure we rebuild the container and it's packages when we build it.

easternbloc:


lennym:
What's the logic behind doing all of those things twice?

lennym:
Would it not be easier just to build it with `--no-cache`?

easternbloc:
wait a mo

easternbloc:
Ignore my previous comments I don't think we'll rebuild this enough to warrant the extra confusion.

easternbloc:
@lennym @joefitter ping!

PR #328, opened 2021-06-21
SD-855 update cookie banner to reflect service

cmedlandho:
# What
A change so that the cookie banner reflects which service the user is on.

# Why
The cookie banner was showing the incorrect service title for the museums and shooting-clubs route

# How 
update to app.js to set the service name depending on the route


shamiluwais (approved):
looks good

PR #334, opened 2021-07-02
SD-996 incorrect servicename on cookie banner

cmedlandho:
# What
An update to show the correct service name for each of the services when viewing the /cookies page

# Why?
When accessing the cookies page the banner was incorrectly showing just the servicename for Prohibited weapons and ammunition licensing instead of the relevant service

# How?
updated app.js to initially set the servicename for the appropriate service and then perpetuate this throughout the form.

# Testing?
Tested locally


shamiluwais (approved):
Looks good

PR #356, opened 2021-09-30
FA-85 and 86: Missing label for invoice address for postcode lookup page and support for uppercase file extensions for document uploaders

SarahJaneLuff:
**Changes**	

Added missing label for invoice address for postcode lookup page.

- invoice-address-select label added to fields.js for all forms.

 Support added for PDF files extensions for file-vault.
    
- file-vault-deployment.yml changed to include support for file extensions in upper and mixed case, via a file-vault update.

### DIFF
--- a/apps/museums/translations/src/en/fields.json
+++ b/apps/museums/translations/src/en/fields.json
@@ -58,6 +58,9 @@
   "invoice-address-manual": {
     "label": "Address"
   },
+  "invoice-address-select": {
+    "label": "Select the address"
+  },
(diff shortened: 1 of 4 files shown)


Alex-Swann (approved):

Alex-Swann (changes requested):
I think update filevault in this PR and leave the whitelisting alone. I think that would be cleaner and it is quick to do

Alex-Swann (approved):

PR #363, opened 2021-12-08
Changes to ensure validation message shows for when a file is over the max file size limit.

SarahJaneLuff:
**Changes**

- maxfilesize changed to 50mb in config.js
- MAX_FILE_SIZE and REQUEST_TIMEOUT added to file-vault-deployment to allow more time and to increase the maximum filesize for the clamAV scan in file-vault.
- Asynchronous problem fixed in existing anf supporting document behaviours, so we wait for a filesize check to complete before calling file-vault. This is to mitigate a 503 error being shown before the validation message is displayed.
- Markdown text changed from 100mb to 50mb for the reduced filesize.
- Added unit tests for supporting and existing-authority document controllers, to test max filesize validation.

### DIFF
--- a/apps/common/controllers/existing-authority-documents.js
+++ b/apps/common/controllers/existing-authority-documents.js
@@ -17,10 +17,10 @@ module.exports = class UploadController extends BaseController {
     super.get(req, res, next);
   }
 
-  process(req, res, next) {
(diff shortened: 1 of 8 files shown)


SarahJaneLuff:
Changes on another PR

PR #371, opened 2021-12-15
Firearms 101: Underlined text present on the firearms forms s5/Museum/Shooting clubs confirm page 

Rhodine:
## What?
Resolved accessibility alert caused by underlined text(Note - this branch is created from [FIREARMS-100](https://github.com/UKHomeOffice/firearms/pull/367), so only the last commit pertains to this pull request).
## Why?
The accessibility alert states that there is underlined non-linked text and suggests that underlines should indicate linked text. The text in this case is in fact linked text and should be underlined but the alert occurs because there is a css style underlining the `<span>` tag surrounding the `<a>` tag.
## How?
Moved css class from span tag to anchor tag. Css styling remains the same but no longer flags the accessibility alert.
## Testing?
Tested locally
## Anything Else?

### DIFF
--- a/apps/common/views/partials/existing-authority-documents-guidance.html
+++ b/apps/common/views/partials/existing-authority-documents-guidance.html
@@ -1,6 +1,6 @@
 <details>
   <summary role="button">
-    <span class="summary">{{#t}}pages.existing-authority.guidance{{/t}}</span>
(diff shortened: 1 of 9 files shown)


shamiluwais (approved):
looks good

shamiluwais (approved):

PR #378, opened 2022-01-14
revert firearms-98: add underline under text

Rhodine:
## What
reverted [Firearms-98](https://github.com/UKHomeOffice/firearms/pull/360/files) so that text is underlined once more.
## Why
To keep it inline with GDS component the text does need to be underlined.
## Testing
tested locally


MaxOSullivan (approved):
Looks good

PR #380, opened 2022-02-08
Add initial acceptance tests 

hjvoid:
### **What?**
Adding acceptance tests to reflect continuity across all HOF forms. 

### **Why?**
Currently not all services under HOF have BDD reporting in place. This makes it difficult for both testers, the businesses and for high level reporting purposes to demonstrate the coverage and health of the acceptance tests on each service.

We also want this to be an automated task, instead of currently requesting coverage from a developer in the form of Drone logs, or 'the build has passed' which isn't giving us the detail we need.

### **How?**
Using combination of Cucumber.js and Playwright.
*Check-email behaviour in the 'supporting-documents' journey has been altered to accept a test email if the env is not set to 'production' in the config file. 

### **Testing?**
Unit test and Lint tests. 

### **Screenshots (optional)**
N/A

### **Anything Else?**
Previous test branch in drone.yml (previously 'test_acceptance_branch') has been renamed 'acceptance_test_branch and replaced with cucumber publish report and playwright setup.
To run the acceptance test, change the `env` variable in the config file to `'ci'`.


config.js
```diff
@@ -7,6 +7,7 @@ const localhost = () => `${process.env.LISTEN_HOST || '0.0.0.0'}:${process.env.P
 module.exports = {
   env: env,
   upload: {
+    skipEmail: env === 'production' ? '' : 'test@test.com',
```
cmedlandho, line 10: Im guessing this means acceptance tests wont work locally unless you have NODE_ENV set locally too? Looks good otherwise!

cmedlandho (approved):
Nice one!

config.js
hjvoid: Yeah I guess so. Would it be better to change it to the same as line 12 `(!env || env === 'ci')` ?

cmedlandho: I think then they wouldnt work in the pipelines? Not certain
I think: 
`skipEmail: env !== 'production' ? 'test@test.com : ' ',`

cmedlandho: Wait that does the same thing lol

hjvoid: Cool! Will change it accordingly and push again - thanks!

PR #387, opened 2022-03-29
Firearms-3 - Added phonenumber validation rules to all phone numbers

MCrawleyHomeOffice:
What?

Lock down phone number fields to only +(0-9)

Why?
The phone number fields currently accept anything

How?
Added the phonenumber validation rule to each field

Testing?
Tested locally

Screenshots (optional)
Anything Else?

### DIFF
--- a/apps/museums/fields/index.js
+++ b/apps/museums/fields/index.js
@@ -15,7 +15,7 @@ module.exports = {
   },
   'contact-phone': {
     mixin: 'input-text',
-    validate: ['required', 'phonenumber']
+    validate: ['required', 'internationalPhoneNumber']
   },
   'same-contact-address': {
     mixin: 'radio-group',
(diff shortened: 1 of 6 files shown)


shamiluwais (approved):

Rhodine (commented):
The error message states "Enter a phone number, like 01632 960 001, 07700 900 982 or +44 808 157 0192" but if you enter 01632 960 001 or 07700 900 982, the same error message appears  - is this because 01632 and 077700 9 numbers are fictitious phone number prefixes? And if so, might it be worth having a message for if the number is invalid?

MCrawleyHomeOffice:
> The error message states "Enter a phone number, like 01632 960 001, 07700 900 982 or +44 808 157 0192" but if you enter 01632 960 001 or 07700 900 982, the same error message appears - is this because 01632 and 077700 9 numbers are fictitious phone number prefixes? And if so, might it be worth having a message for if the number is invalid?

Many thanks Rhodine, I will look into this.

PR #399, opened 2022-04-19
Added new root page and relevant navigation

MCrawleyHomeOffice:
What?
No root page exists meaning locals fail to load. This results in a 500 error in production when you click Continue on the root privacy page. We have now added a root page with 4 options so the user can choose their journey.

Why?
This gives a better user experience and less likely to experience unwanted journeys.

How?
Added a choose-a-journey radio field and the appropriate forking.

Testing?
Tested manually.

Screenshots (optional)
Anything Else?

### DIFF
--- a/apps/common/fields/index.js
+++ b/apps/common/fields/index.js
@@ -1,6 +1,19 @@
 'use strict';
 
 module.exports = {
+  'choose-a-journey': {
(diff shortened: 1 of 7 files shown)


Rhodine (approved):
lgtm

PR #400, opened 2022-04-20
FIREARMS-X - Removed so-acceptance

MCrawleyHomeOffice:
What?
Vulnerability found in so-acceptance has highlighted a need to remove/update/replace the async npm package. After discussing with the team, this is no longer required and can be removed as we have now migrated to playwright.

Why?
No longer required as we've migrated to playwright.

How?
Remove so-acceptance package.

Testing?
New acceptance tests are running successfully

Screenshots (optional)
Anything Else?


Rhodine (approved):
lg

PR #401, opened 2022-04-25
FIREARMS-113 Fixed issue with supporting doc ref check and added test

MCrawleyHomeOffice:
What?
Current icasework ref check is failing due to a minor bug in the logic

Why?
The check was not comparing with the user input

How?
Added string comparison

Testing?
Tested locally and added additional acceptance test

Screenshots (optional)
Anything Else?


hjvoid (approved):
Looks good to me!

PR #423, opened 2022-07-08
Add auth files and update auth logic

Alex-Swann:


apps/common/behaviours/casework-submission.js
```diff
@@ -6,10 +6,9 @@ const client = new StatsD();
 
 const Compose = func => superclass => class extends superclass {
   prepare() {
-    if (typeof func === 'function') {
```
Alex-Swann, line 9: func is always a function as the 4 journeys this Compose function is called in the submission process, a function which builds the payload to send to iCasework is called here

apps/common/behaviours/casework-submission.js
```diff
@@ -6,10 +6,9 @@ const client = new StatsD();
...
-      return Object.assign(super.prepare(), func(this.toJSON()));
-    }
-    return super.prepare();
```
Alex-Swann, line 12: therefore this line is superfluous

apps/common/behaviours/casework-submission.js
```diff
@@ -6,10 +6,9 @@ const client = new StatsD();
...
+    return this.auth().then(token => {
```
Alex-Swann, line 9: This prepare step essentially overrides the one that exists in the HOF model. Superclass declared above is essentially either the `apps/common/models/i-casework.js` or `apps/common/models/i-casework-documents.js`; the latter which extends off the former anyway. The i-casework model has now been updated to extend off the Auth model and HOF models allowing us to call this.auth() here, generate a bearer token and pass this both to the functions that build the payload that is sent to icasework, but also the super.prepare() step which calls the final pdf application, allowing us to update all upload urls with the token bearer at once.

P.s. I think this process is overly complicated as this took me ages to work out. I want to gradually refactor this into something far more sensible. This whole override BS is completely unnecessary imho and I reckon we can simplify it for others.

apps/common/models/file-upload.js
```diff
@@ -3,10 +3,10 @@
 
 const url = require('url');
 
-const Model = require('hof').model;
+const AuthToken = require('./auth-token');
 const config = require('../../../config');
 
-module.exports = class UploadModel extends Model {
+module.exports = class UploadModel extends AuthToken {
```
Alex-Swann, line 9: Little did I know that the HOF Model called which this extends off, calls the auth() function below for POSTing the uploads in the first place even though it isn't explicit in the service code. Eventually I worked that out. In this case we need to share the auth logic with this upload component but I've separated it out so other processes can extend off the AuthToken model but where it's functionality is still accessible here for the HOF model process.

apps/common/models/file-upload.js
```diff
@@ -23,49 +23,14 @@ module.exports = class UploadModel extends Model {
...
+      // thus needs to be declared in the parent class which extends off the HOF model too.
       this.request(reqConf, (err, data) => {
         if (err) {
           return reject(err);
         }
         return resolve(data);
       });
-    }).then(data => {
-      return this.auth().then(bearer => {
-        data.url = (data.url.replace('/file', '/vault')) + '&token=' + bearer.bearer;
```
Alex-Swann, line 34: We don't augment any upload url with the token bearer here anymore. We leave it right to then end when the user clicks continue on the Declaration page which is exactly when the payload is submitted to icasework. This means we both generate a token bearer, add it to all the urls and pass it on to iCW without any delay by the user. This means we can reduce the JWT token expiry from 1-2 hours, to literally a few minutes in the Keycloak realm. Also means any failure to get the token in the first place will trigger an error for the user before anything submits, and means we know no latency on their part will cause an expiry issue and thus the duplication problems we've seen on iCW. It should sort all of it out.

apps/common/models/i-casework-getcase.js
```diff
@@ -15,10 +15,11 @@ module.exports = class DocumentModel extends Model {
     return crypto.createHash('md5').update(this.get('reference-number') + date + config.icasework.secret).digest('hex');
   }
 
-  prepare(data) {
-    const props = super.prepare(data);
-    props.CaseId = this.get('reference-number');
-    return props;
+  prepare() {
+    return super.prepare().then(props => {
```
Alex-Swann, line 19: Because of the Compose function above and needed to fetch a token bearer, it means that prepare step has now become async. As it overrides any other prepare steps declared in other models to be on the safe side, and match the pattern in the HOF model, all declared prepare steps have been handled to be async. So here we assume super.prepare is async...

apps/common/models/i-casework-getcase.js
```diff
@@ -29,9 +30,11 @@ module.exports = class DocumentModel extends Model {
   }
 
   fetch() {
-    const options = this.requestConfig({});
-    options.qs = this.prepare();
-    options.method = 'GET';
-    return this.request(options);
+    return this.prepare().then(qs => {
```
Alex-Swann, line 33: ...and thus we assume calling this.prepare, because of super.prepare possibly being async is handled thus here.

apps/common/models/i-casework-getcase.js
```diff
@@ -29,9 +30,11 @@ module.exports = class DocumentModel extends Model {
...
+      const options = this.requestConfig({});
+      options.qs = qs;
+      options.method = 'GET';
+      return this.request(options);
```
Alex-Swann, line 37: because this.request is returned and is an async func already we don't need to adjust anything handling this fetch function as it was treated as async to begin with

apps/common/models/i-casework.js
```diff
@@ -1,11 +1,11 @@
 'use strict';
 
-const Model = require('hof').model;
+const AuthToken = require('./auth-token');
 const crypto = require('crypto');
 
 const config = require('../../../config');
 
-module.exports = class CaseworkModel extends Model {
+module.exports = class CaseworkModel extends AuthToken {
```
Alex-Swann, line 8: By extending off the auth model, it means this.auth() works in the Compose function above because any superclass declared there essentially are either this model or a descendant of it. If we used the HOF model directly, this.auth() doesn't do anything without a different setup.

apps/common/models/i-casework.js
```diff
@@ -15,7 +15,7 @@ module.exports = class CaseworkModel extends Model {
     return config.icasework.url + config.icasework.createpath;
   }
 
-  prepare() {
+  prepare(token) {
```
Alex-Swann, line 18: Only in the case of submission will token be defined as it is passed to the prepare step by the Compose function. If we have missed anything for and scenarios where pdf-upload is being grabbed (unlikely), token bearer below will cause a ReferenceError in the application. If we test all journeys and check the unit tests I imagine we should be alright. Again this would be simpler once we refactor out this whole Compose/prepare override gubbins.

apps/museums/models/submission.js
```diff
@@ -2,7 +2,7 @@
 
 const _ = require('lodash');
 
-module.exports = data => {
+module.exports = (data, token) => {
```
Alex-Swann, line 5: Any `apps/*/models/submission.js` files in all 4 journeys are essentially a function that builds the iCW payload for that particular journey. Now I'm passing the bearer token to each of them as a second parameter we can update all doc urls here instead meaning we can generate the token at the very last minute.

apps/common/models/i-casework.js
```diff
@@ -41,16 +40,19 @@ module.exports = class CaseworkModel extends Model {
...
-    options.method = 'POST';
-    if (!config.icasework.secret || !config.icasework.key && config.env !== 'production') {
-      return Promise.resolve({
-        createcaseresponse: {
-          caseid: 'mock caseid'
-        }
-      });
-    }
-    return this.request(options);
+    return this.prepare().then(formData => {
```
Alex-Swann, line 43: Handling the prepare step as async as I said above. Notice there is no token being passed to it. That is because the pdf-upload logic in the prepare step above isn't used if the step itself isn't being overridden by the prepare step in the Compose function. If it is, then this.prepare() calls that first, the token is generated and then passed to the prepare step above.....I told you it was confusing LOL

apps/common/behaviours/casework-submission.js
```diff
@@ -1,15 +1,16 @@
 'use strict';
 
+const AuthToken = require('../models/auth-token');
 const CaseworkModel = require('../models/i-casework');
 const StatsD = require('hot-shots');
 const client = new StatsD();
 
 const Compose = func => superclass => class extends superclass {
   prepare() {
-    if (typeof func === 'function') {
```
Alex-Swann, line 9: func is always a function which is essentially creates a payload object to send to icasework on all 4 journeys.

apps/common/behaviours/casework-submission.js
```diff
@@ -1,15 +1,16 @@
...
-      return Object.assign(super.prepare(), func(this.toJSON()));
-    }
-    return super.prepare();
+    const model = new AuthToken();
+    return model.auth().then(token => {
```
Alex-Swann, line 11: Prepare statement is now async, so we have to handle other prepare statements calls elsewhere in the service as if they are async too just in case. This is also how this step is treated in the HOF model anyway. Here we generate a token bearer and pass it to the payload builder functions and the prepare step essentially in the `apps/common/models/i-casework.js` model where we can update all urls at once.

apps/common/behaviours/casework-submission.js
```diff
@@ -1,16 +1,19 @@
...
 const CaseworkModel = require('../models/i-casework');
 const StatsD = require('hot-shots');
 const client = new StatsD();
 
 const Compose = func => superclass => class extends superclass {
   prepare() {
     if (typeof func === 'function') {
-      return Object.assign(super.prepare(), func(this.toJSON()));
+      const model = new AuthToken();
+      return model.auth().then(token => {
```
Alex-Swann, line 12: Generate auth token here and pass to i-casework model prepare step and payload builders that setup the object that is sent to icasework

apps/common/models/i-casework.js
```diff
@@ -41,16 +40,19 @@ module.exports = class CaseworkModel extends Model {
...
-    options.method = 'POST';
-    if (!config.icasework.secret || !config.icasework.key && config.env !== 'production') {
-      return Promise.resolve({
-        createcaseresponse: {
-          caseid: 'mock caseid'
-        }
-      });
-    }
-    return this.request(options);
+    return Promise.resolve(this.prepare()).then(formData => {
```
Alex-Swann, line 43: this.prepare is usually sync, however the Compose function/model above returns an async prepare step if needed to generate an auth token. This handles that by always resolving a promise whether it is sync or async.

MCrawleyHomeOffice (approved):
lgtm

PR #425, opened 2022-07-19
Revert "Add auth files and update auth logic"

MCrawleyHomeOffice:
Reverts UKHomeOffice/firearms#423


MCrawleyHomeOffice:
Created revert for recent auth changes in case we decide this is the cause of the current 500 issues

PR #427, opened 2022-07-19
Adjusted file upload desc length to icasework max

MCrawleyHomeOffice:
Hotfix to fix the 500 internal server errors we are getting in icasework


cmedlandho (approved):
Looks good - Nice work

PR #428, opened 2022-07-25
Minor change to force build for GA4 update

MCrawleyHomeOffice:
Made a minor change to the README to force a Drone build so we can use the latest hof-config settings


shamiluwais (approved):

MCrawleyHomeOffice:
This may no longer be required.

PR #453, opened 2023-02-20
Updated ingress config due to breaking change

MCrawleyHomeOffice:
Updated ingress config due to breaking change


shamiluwais (approved):
lgtm

PR #470, opened 2023-08-18
FIR-177: Import page uses selection dropdown for countries

ShahSheel:


ShahSheel:
A further ticket will be raised for Firearms to address valdation for _all_ fields using countries/lookaheads

Rhodine-orleans-lindsay (approved):
lgtm

PR #493, opened 2024-07-04
HOFF-794: Deploy Internal Ingress for performance testing

adityababumallisettiHO:
### **WHAT?**

Testers require Internal Ingress for automation testing. UAT Environment of Firearms never had Internal Ingress. We are deploying internal ingress . We are not allowing public access to any of the URLs. We also need to set Ingress class to nginx-internal. 
https://collaboration.homeoffice.gov.uk/jira/browse/HOFF-794

### **WHY?**

External ingress is returning 403 Forbidden error. 

### **HOW?**

Update if condition to Ingress annotations which are exported to internal ingress file in firearms repo

### **Testing?**

We can test by deploying to UAT. If we dont have admission policy in the namespace we will see the Drone build complaining about it while deploying to UAT

### **Anything Else?**

### DIFF
--- a/.snyk
+++ b/.snyk
@@ -50,4 +50,24 @@ ignore:
     - busboy-body-parser > busboy > dicer:
         reason: No update available
         expires: '2024-04-10T12:21:04.350Z'
+  SNYK-JS-AXIOS-6032459:
+    - hof > notifications-node-client > axios:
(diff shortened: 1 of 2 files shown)


jamiecarterHO (approved):
lgtm

TemitopeAyokuHO (approved):

